Issue #136, opened 2021-03-23
Harvest Report in Barn Kit Tab


braughtg:
See Harvest -> Report in the original FarmData at [https://farmdata.dickinson.edu/guest.php](https://farmdata.dickinson.edu/guest.php)

User friendly query for harvest data.    Choose date range, field and crop, get harvest data for that   period.  ALL is allowed value for field   and crop to provide summary data, or drill down into specific fields.  Data report includes total yield expressed   in the default harvest unit for crop selected (FD converts harvest data   inputs from a variety of units to crop default unit, based on unit conversion   table).    Report also shows   "yield per row foot" which is calculated by dividing total yield   for the period by total row feet planted (from transplant or direct seed   tables) for that crop in selected fields over the same period.   Report includes labor data in units per   labor hour

Batese2001:
Currently as merged with main according to pull request #155:
- When one visits the harvest report tab, one is able to type a title, select a date range (the default of which is from the start of the current year to the current date), field and crop from the database. 
- All is an allowed value for both crop and field. 
- Results are produced when one clicks the generate report button. 
- Each harvest log in the report contains the date it was harvested, the crop, the field, the individual yield in the given unit, the worker who inputted the log, and a semi-functional delete button. 
- Below this table there is a yield table that consists of all the crops alongside their respective total yields. 

With the addition of pull request #158:
- Below this table is a download link that converts the two tables into one csv file containing all of the data, which can be downloaded and named after the inputted title. 

Problems:

1. The fields and crops in the dropdown are found directly in the database. However, if it is entered incorrectly or is not updated in an input form and thus makes its way into the log database, then it cannot be filtered to by the selection and will only be accessible via the ALL option. Thus there should be a way of preventing this or otherwise updating the database to include these other options.

2. Some logs do not have necessary information such as a harvest unit or a crop name. This makes the code crash. This should be handled or otherwise prevented higher up. 

3. Results are limited to 100 logs for each search due to the limitations of the request. In the data returned by the request, there is information including the current result page, the first and the last, as well as the next. This should be used to create a bigger and more accurate table if a result could be more than 100 logs. 

4. Crops occasionally are inputted with different units, causing them to be unable to be added together accurately. In this code, it adds them separately but I am unsure if this is the desired result. 

5. The delete button does not update the database and instead only updates the lists on the page, giving the correct appearance but leaving this as a potential error when others wish to get the same report. 

[...]

PR #155, opened 2021-04-19
Merged Report

Batese2001:
__Pull Request Description__

For Issue #136

User Story: Field Selection

- Create the field selection
- Connect the field to API request and get back the data from the chosen field
- Show the harvest logs from the chosen field in a table

User Story: Report Date Range Selection 

- Display start date and end date label
- Show two separate field to choose date next to the corresponding labels
- Default start date is the beginning of current calendar year, default end date is the current day
- Filter the harvest logs by date

This will add total yield to the Report.

It should:

- Compute the yield at the end of the report
- Use default harvest unit
- Separately add same crops together 

This will add crop selection to the Report.

It should:

- Add a crop dropdown
- Filter results by crop
- Ignore crop using ALL filter

This will add the basics of a Harvest Report generator to the Barn Kit tab.

It should:

- Generate a table using an button
- Get data from the database
- Have the ability to change the name of the Report
- Have instructions and a title for the tab

---
__Licensing Certification__


braughtg (changes requested):
This is looking good. I would be good to have the min and max properties of the dates bound so that the start date must be before the end date.  Make that change and request another review and we'll be good to merge this.

braughtg (approved):
I made one additional change.  I also bound the end date to be after the start date.

PR #180, opened 2021-06-11
Date selection components

IrisSC:
__Pull Request Description__

Adds the dateSelection and dateRangeSelection Components to the resources file in fd2_tabs. The date selection is the child component, it allows the user to selected a date. The programs are able to put in props that will effect the earliest, lasted, and default date. The dateRangeSelection component has two dateSelection components and has it so the user can select a start and end date. Developers using this are able to input a default start and end date in the html page. This also includes the cypress tests for these two components, that test user input of these components. The cypress tests is located in the same file.

---
__Licensing Certification__


farmdata2_modules/fd2_tabs/fd2_example/cypress/integration/componentTesting.spec.js
```diff
@@ -0,0 +1,13 @@
...
+          cy.visit('/farm/fd2-example/componentTesting')
+  })
+  context('Crop Dropdown', () => {
+    it('accepts input from search bar', () => {
+      cy.get('crop-picker-element')
+        .type('lettuce')
+        .should('have.value', 'lettuce')
+    })
+  })
+})
```
braughtg, line 13: Looks like this is left from the preliminary testing.  If so it should be reomved.

farmdata2_modules/fd2_tabs/fd2_example/testComponents.html
```diff
@@ -0,0 +1 @@
+!!!
```
braughtg, line 1: This file appears to be from the preliminary testing and should be removed if so.

braughtg (changes requested):

braughtg:
This PR was superseded by #183 and the relevant changes from this PR were included in that one.

PR #197, opened 2021-06-22
Delete function

josieecook:
__Pull Request Description__

In progress cypress tests for the delete function

---
__Licensing Certification__


braughtg:
The following seems to working using the cy.request in place of an axios call to create the logs.  Using cy.request for things that are test-only seems to be the preferred approach.

```
    context('delete function', () => {
        it.only('deletes a log based on log ID', () => {
            getSessionToken()
            .then(function(token) {
                console.log('making log')

                req = {
                    url: '/log.json?type=farm_observation',
                    method: 'POST',
                    headers: {
                        'X-Requested-With': 'XMLHttpRequest',
                        'X-CSRF-TOKEN' : token,
                    },
                    body: {
                        "name": "heyyy",
                        "type": "farm_observation",
                        "timestamp": "1526584271",
                    }
                }
                cy.request(req).as('created')
                
                logID = -1
                cy.get('@created').should((response) => {
                    expect(response.status).to.equal(201)
                    logID = response.body.id
                })

                deleteLog('/log.json?type=farm_observation', logID, token)
            })
            .then(function(response) {
                
                console.log("delete done")
            })
        })
    })
```

PR #198, opened 2021-06-24
Delete Function and Test

josieecook:
__Pull Request Description__

A functional deleteLog function and functional but messy cypress tests for that function.

---
__Licensing Certification__


braughtg:
Here is a structure that seems to be effective for the delete test and should generalize to the other tests as well.

```
        it('deletes a log based on log ID', () => {
            getSessionToken()
            .then(function(token) {
                console.log('making log')

                req = {
                    url: '/log.json?type=farm_observation',
                    method: 'POST',
                    headers: {
                        'X-Requested-With': 'XMLHttpRequest',
                        'X-CSRF-TOKEN' : token,
                    },
                    body: {
                        "name": "pleasegoaway",
                        "type": "farm_observation",
                        "timestamp": "1526584271",
                    }
                }

                logID = -1
                cy.request(req).as('created')
                cy.get('@created').should((response) => {
                    expect(response.status).to.equal(201)
                    logID = response.body.id
                })

                cy.wrap(deleteLog('/log.json?type=farm_observation', logID, token)).as('delete')
                cy.get('@delete').should((response) => {
                    expect(response.status).to.equal(200)
                })

                url = '/log.json?type=farm_observation&id=' + logID
                cy.request(url).as('check')
                cy.get('@check').should((response) => {
                    expect(response.body.list.length).to.equal(0)
                })
            })
        })
```

braughtg:
Note that the tests for getAllPages can also be updated to use this.  It may also remove the need for the hacky wait(1) that is in there.  So it would be worth checking to see if that can be taken out.

braughtg:
Superseded by #201.

PR #204, opened 2021-06-25
Conflicting Cypress Tests: createLog and deleteLog

josieecook:
__Pull Request Description__

The cypress tests for createLog and deleteLog run successfully when separate but will not fully run/fail when run in the same file.

---
__Licensing Certification__


braughtg:
The problematic tests were merged and issue #210 was created to address the need for the tests to be refined.

PR #227, opened 2021-07-01
Table Component - Strings to HTML

josieecook:
__Pull Request Description__

The table component can now accept strings containing html elements and render those elements in the table. The cypress tests have been updated to accommodate this change, and the array passed to the table in ex1.html has been altered to demonstrate it.

---
__Licensing Certification__


farmdata2_modules/fd2_tabs/resources/CustomTableComponent.spec.comp.js
```diff
@@ -66,7 +66,7 @@ describe('custom table component', () => {
         it('prop change updates table', () => {
             prop.rows[0].data = [ 5, 10, 'Wahooo' ]
 
-            cy.get(':nth-child(1) > :nth-child(3) > p')
+            cy.get(':nth-child(1) > :nth-child(3) > div')
                 .should('have.text', 'Wahooo')
         })
     })
```
braughtg, line 72: Seems like this test should somehow check that the html entered was rendered and not displayed as plain text.  So maybe a test that has html in the cell with a <p> tag or something and then checks that the text does not include the tag?

Also using data-cy= attributes on the elements is the preferred way to access them as opposed to using the nth-child etc... as that position may change if the data in the table is changed.  Here I'd anticipate using the row and column indices as part fo the data-cy value so you can pick out the cell directly.

PR #233, opened 2021-07-08
Seeding Report Page Tests

josieecook:
__Pull Request Description__

The Cypress tests for the seeding report page cause a 403 error when making API requests.

---
__Licensing Certification__


braughtg:
In your tests you will need to log in and visit the tab for each "it"  The login credentials and session key are reset before each "it".  So you can do that with a beforeEach if you will be using the same user for all "its" or you can add the login and visit to each "it" separately.

PR #245, opened 2021-07-13
Log type map

IrisSC:
__Pull Request Description__

<Adds mapping functions for getting log category to id and vice versa as well as cypress tests for the new functions. Also fixes wrong ids in other map function tests.>

---
__Licensing Certification__


farmdata2_modules/fd2_tabs/resources/FarmOSAPI.spec.js
```diff
@@ -98,7 +100,7 @@ describe('API Request Function', () => {
         })
     })
     
-    context('test maping functions', () => {
+    context.only('test maping functions', () => {
```
braughtg, line 103: The .only should be removed here so that all tests will run in the version in the upststream.

IrisSC: I have removed the .only. Thank you for pointing it out.

PR #247, opened 2021-07-14
Table Editing Changes

josieecook:
__Pull Request Description__

Added a "middle-man" object to the table component that saves the edited row data to be compared to the original and then emitted.

---
__Licensing Certification__


farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
```diff
@@ -62,38 +62,52 @@ let CustomTableComponent = {
...
     methods: {
         editRow: function(index){
-            this.rowToEdit = index
+            this.rowToEditIndex = index
+            this.editedRowData = { 
+                'id': this.rows[index].id,
+                'data': this.rows[index].data
+            }
+
+            console.log(this.editedRowData)
```
braughtg, line 79: Please remove console.log statements from the production code.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
```diff
@@ -62,38 +62,52 @@ let CustomTableComponent = {
...
             for(i=0; i < this.indexesToChange.length; i ++){
                 let key = this.headers[this.indexesToChange[i]]
-                jsonObject[key] = row.data[this.indexesToChange[i]]
+                jsonObject[key] = editedRowData.data[this.indexesToChange[i]]
             }
             this.indexesToChange = []
 
+            this.editedRowData = {}
+
+            console.log(jsonObject)
```
braughtg, line 93: Please remove console.log statements from the production code.

braughtg (changes requested):

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
```diff
@@ -62,38 +62,48 @@ let CustomTableComponent = {
...
+
             this.$emit('row-edited', jsonObject, id)
         },
         deleteRow: function(id){
             this.$emit('row-deleted', id)
         },
         changedCell: function(itemIndex){
             if(!this.indexesToChange.includes(itemIndex)){
-                this.indexesToChange.push(itemIndex)
+                if(this.editedRowData.data[itemIndex] != this.rows[this.rowToEditIndex].data[itemIndex]) {
```
braughtg, line 98: Would it make sense to do this check at the end as well?  That way if something is changed and then changed back to what it was originally it will not be sent as a change?

braughtg (commented):
Please, think about and respond to the question in the comment on the changes.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
josieecook: What about adding an else if in this method that checks if the index is already in indexesToChange? Then we can remove it here rather than doing it later on.

braughtg: Yes, I think this function could be restructured to add changes and to also remove indices and data that were "unchanged" (i.e. changed and then changed back).

braughtg (approved):

josieecook:
Please wait to merge until the cypress tests are updated.

Issue #273, opened 2021-07-23
Seeding Input: restructure order of Inputs


IrisSC:
The Data section should be broken up into two sections, with the date, crop, comments, and area in the first section. The second section should have everything else and be after the Tray/Direct Seeding section. The second section should be called something other than data.

![reorderingDataInputs](https://user-images.githubusercontent.com/54850069/126838199-a64fda69-909c-4c40-902a-c0ed3a6990e6.png)

To test this go to the Seedings Input subtab under FieldKit tab, which is on the home page.  

The code can be found by going to farmdata2_modules/fd2_tabs/fd2_field_kit/seedingInput.html file.

braughtg:
It appeared that some of this reordering was in PR #263 as well.  There is still work to be done here as described in the discussion on PR #281, so I'll leave this open until that is closed.

PR #281, opened 2021-09-17
Reorder Seeding  Inputs

IrisSC:
__Pull Request Description__

A new category of inputs was created called "Labor" and the number of workers & time spent was moved down into this category. The cypress tests were also changed so that they would work with the new ordering.

Closed: #273 
---
__Licensing Certification__


braughtg:
I like this change.  Could we also get all fields in the labor box onto one line?  Try:
* Using just "Workers" as the label  for the first field and making that data entry box smaller (enough for 2 digits).
* Shortening "Time Spent" to "Time:"
* Eliminating the label "Units:" and just putting the minutes/hours dropdown next to the Time box.

IrisSC:
The first two changes I can do easily. I have a question, for the third. When I eliminate the word 'units' from the before the dropdown, the dropdown component will change it to the default, which is 'Select'. Do you want me to leave it saying 'Select' and create an issue to change the dropdown component?

IrisSC:
While, I agree with making the labels of the inputs shorter. After putting them all on the same line, I am not a big fan of how it looks, even if I were to get them centered. They still wouldn't seem to fit with the rest of the page, at least to me.

image

braughtg:
I think this looks good and may be quite functional on a small device because it would require less scrolling. Centering would be good. 

Though If you have an alternative design idea maybe make another branch and we could A/B test them with Matt. 

Also I just noticed a typo in the tray seeding section. Plants should be “planted” on the “Number of Seeds …” input.

braughtg:
I merged #286 that added the * to the required fields.  That generated conflicts with this PR.  Please resolve the conflicts and I will merge this as well.

PR #282, opened 2021-09-21
Spinner reappear

### LINKED ISSUE #270: Seeding Report - Cypress Tests for Loading Spinner
Tests should be added to the seeding report page cypress test file for the loading spinner. These tests would ensure that it appears and disappears when appropriate.

Should test that:
* The spinner is shown on the page when an API call has been made
* The spinner remains on the page until the whole response has been returned
* The spinner disappears once the whole response has been returned
* The spinner reappears when new dates are selected and a new API call is made. (See Issue #272)

### LINKED ISSUE #272: Seeding Report - Loading spinner should reappear when new dates selected
The loading spinner currently only appears during the first API call after the seeding report page loads. It does not reappear whenever new dates are selected and a new API call is made.

This is the spinner with the desired behavior:
spinner

The seeding report page should be edited so that the spinner appears whenever any new API call is made, not just the first call.

Cypress tests should be added for this behavior, see Issue #270

Batese2001:
__Pull Request Description__

The loading spinner that appears when the "Generate Report" button is clicked, now reappears when it is clicked more times. 

Closes #272 
Closes #270 

---
__Licensing Certification__

### DIFF
--- a/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
+++ b/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
@@ -56,7 +56,7 @@ describe('Testing for the seeding report page', () => {
 
         
     })
-    context.only('can see spinner at appropriate times', () => {
+    context('can see spinner at appropriate times', () => {
         before(() => {
             cy.login('manager1', 'farmdata2')
             cy.visit('/farm/fd2-barn-kit/seedingReport')
--- a/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
+++ b/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
@@ -83,22 +83,6 @@ describe('Testing for the seeding report page', () => {
                 cy.get('[data-cy=loader]').should('not.exist')
             }) 
         })
-        it.only('spinner remains before API return', () => {
-            cy.get('[data-cy=start-date-select]')
-                .type('2019-01-01')
-            cy.get('[data-cy=end-date-select]')
-                .type('2019-05-15')
-            cy.get('[data-cy=generate-rpt-btn]').click()
-            cy.get('[data-cy=report-table]', { timeout: 20000 }).find('tr').its('length').then(length =>{
-                expect(length).to.equal(101)
-                cy.get('[data-cy=loader]').should('exist')
-            })
-            cy.wait(10000)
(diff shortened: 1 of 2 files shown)


farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
```diff
@@ -56,6 +56,58 @@ describe('Testing for the seeding report page', () => {
...
+                .type('2019-01-01')
+            cy.get('[data-cy=end-date-select]')
+                .type('2019-05-15')
+            cy.get('[data-cy=generate-rpt-btn]').click()
+            cy.get('[data-cy=report-table]', { timeout: 20000 }).find('tr').its('length').then(length =>{
+                expect(length).to.equal(101)
+                cy.get('[data-cy=loader]').should('exist')
+            })
+            cy.wait(10000)
+            cy.get('[data-cy=report-table]').find('tr').its('length').then(length =>{
```
braughtg, line 97: Would setting a timeout here work instead of using the wait? I think it would be less brittle with respect to running on different machines.

Batese2001: I have not been able to use it. The problem is that with a 2 page request cypress sends the request, gets the request back, finds the table, gets the second page of the request, finds the row number and then checks it. This means that I am actually checking the row number of the wrong number of requests. I have gotten around this somewhat by sending a 3 page request and checking the second page rather than the first so that I can check what I was checking before. There must be a better way, as I have noticed inconsistent results on my own machine, perhaps due to API calls taking different amounts of time each time.

braughtg: Okay.  Let's just drop this test.  We have one that checks that the spinner appears when the report is being generated and we have one that checks that it is gone when the results are displayed.  I think that will be sufficient and would flag the most likely issues if something chagnes.

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
```diff
@@ -56,6 +56,42 @@ describe('Testing for the seeding report page', () => {
 
         
     })
+    context.only('can see spinner at appropriate times', () => {
```
braughtg, line 59: Using .only is great for development and testing of our changes. Please remove the .only here and rerun all tests to ensure they still pass.  The commit the version without the .only so that all tests will run in the upstream version.

Issue #284, opened 2021-09-22
Hide Seeding Report Summary Tables While Loading


braughtg:
The summary tables that appear at the bottom of the Seeding Report should appear only after all of the data has been loaded.  So while the data is loading the spinner should be the last element on the page. Once the data is loaded the spinner should disappear (as it currently does) and the summary table or tables should appear (related to #259).

Issue #259, opened 2021-07-23
Seedings Report: message in Tray/Direct Summaries


IrisSC:
A user may have all selected for Type of Seeding, but may have selected a crop or area that only has Tray or Direct Seedings. When this happens the other summary will show 0 or NaN (not a number), instead have that summary show a message: “There are no [Tray or Direct] Seedings in the options you have chosen.”

To test this go to the Seedings Report subtab under BarnKit tab, which is on the home page.  Selected a date range, and try filtering different Areas and Crops.

An example is shown below:
![NeedMessageTrayDirectSummaries](https://user-images.githubusercontent.com/54850069/126825600-ffa1628e-cd1a-4240-9634-1fdf9b7d8904.png)

This issue is complete when the message appears in Tray and Direct seeding summaries, when there are no current logs in the respective summaries.

braughtg:
Also add Cypress tests that confirm this behavior.

braughtg:
Instead of a message, if there are only tray seedings or only direct seedings the other summary table should be hidden.

braughtg:
This is related to #284 and #258.  A single PR that solves all three would be welcome.

Batese2001:
Just for clarifications sake, you wanted me to have the summary table with no corresponding logs to show up with a message, so that the user is not confused, correct?

braughtg:
> Just for clarifications sake, you wanted me to have the summary table with no corresponding logs to show up with a message, so that the user is not confused, correct?

I think the desired behavior would be:
- If user filters for just tray or just direct seedings then only one table appears.
- if user filters for both, and the results contain both then both tables appear.
- if user filters for both, and the results contain only direct or only tray (because thats what exists in the date range) then both tables appear, but the one with no records contains a message.

If you have other thoughts on what might be better from a usability perspective I'm open to suggestions.

Issue #258, opened 2021-07-23
Seeding Report: no logs in table, display a message


IrisSC:
If there are no logs in the table then the filters, table, and summaries do not show up. The page where they should be is just blank. Instead we want there to be a message saying “There are no logs in the time range you have selected. Please select new dates.”

There are two scenarios when the filters, table, and summaries disappear. First is when there are no logs within the date range, the second is when the user deletes all the logs in the date range. This is likely due to there being no logs in the array seedingLogs. 

To test this go to the Seedings Report subtab under BarnKit tab, which is on the home page.

![NoLogsNeedMessageIssue](https://user-images.githubusercontent.com/54850069/126822208-723909d4-2f11-464e-b110-aea9d3ad74a5.png)

braughtg:
Also include Cypress tests that confirm the behavior.

Batese2001:
I will start work on this

Batese2001:
In order for the error message to display appropriately, that is, displaying once the results have been gathered and not while waiting for results to load, issue #272 should be solved first

braughtg:
This is also related to #284.  It might be easy to combine them and make a PR that solves both.

PR #289, opened 2021-10-05
Summary Table Timing and No Log Messages

Batese2001:
__Pull Request Description__

- Summary Tables now show up after the report table is completely loaded to avoid confusion. Issue #284 
- When filters leave only Direct or Tray Seeding logs in the table, a message is shown in the opposite table to notify the user of this. Issue #259
- When a date range is selected with no logs, a message is displayed instead of the table. Issue #258

---
__Licensing Certification__

### DIFF
--- a/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
+++ b/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
@@ -93,6 +93,40 @@ describe('Testing for the seeding report page', () => {
         })
     })
 
+    context('shows message when only one type', () => {
+        before(() => {
+            cy.login('manager1', 'farmdata2')
+            cy.visit('/farm/fd2-barn-kit/seedingReport')
+
+        })
+
+        it('show both tables with two types', () => {
+            cy.get('[data-cy=start-date-select]')
+                .type('2020-05-01')
+            cy.get('[data-cy=end-date-select]')
+                .type('2020-06-01')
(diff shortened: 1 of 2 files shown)


braughtg:
A hand test looks good on this.  However, I am having trouble getting the automated tests to run reliably.  It generally seems that they run okay if only one of them is run at a time.  However, when the full suite is run there are failures (often 403 status code - unauthorized).  I am suspecting this may have to do with an interaction between the `beforeEach` on the full `describe` block and the `before` in the individual `context` blocks.  They both do a login.  Perhaps it should only be in one place or another?  Maybe just login in the `beforeEach` in the `describe` and then `visit` the appropriate page in the `context`?  

Try a few things and see where you get.  The goal will be to be able to run the full suite of tests at once with no failures.

Batese2001:
I noticed while testing the the tests ran but some seemed to fail for unrelated reasons in other tests I was not responsible for. Should I try to figure out what those were testing and fix that too?

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
```diff
@@ -93,6 +92,112 @@ describe('Testing for the seeding report page', () => {
         })
     })
 
+        context('can see No Logs message at appropriate times', () => {
+        before(() => {
+            //cy.login('manager1', 'farmdata2')
+            //cy.visit('/farm/fd2-barn-kit/seedingReport')
```
braughtg, line 98: I think you want the visit here but not the login.  Also should these be a beforeEach or a before?

PR #305, opened 2021-10-20
Cache Area and Crop Arrays

### LINKED ISSUE #278: Cache Area/Crop Lists for Seeding Inputs
Explore ways to possibly cache the area and crop lists for the seeing input form.  It takes too long to retrieve these from the server every time the form is accessed.  
* It may be possible to improve the performance of this on the FarmOS/Drupal end.  This would require some research into the general performance of the API.  A search of the FarmOS forums and maybe a conversation with the devs on their chat platform would be a good place to start.  If this approach is viable then it would probably be the preferred approach since it would require no changes to the front end and would always ensure the most up to date data is available.
* Another possibility would be to retrieve and cache these lists the first time they are needed. This could be done in one of the Vue.js lifecycle hooks.  That hook could check if the needed values had been cached and if not make the API requests and cache the results. Then the dropdowns could be populated more quickly from the cached results.  The big question here is where those lists could be stored so that they persist across changes in tabs and sub-tabs, etc.  Again a question to the FarmOS community explaining that we'd like JS code running in a tab to be able to cache information so that it is available later, (i.e. after the tab is changed and then changed back).

IrisSC:
__Pull Request Description__

Previously each time the user opened the seeding input, the crop and area list would take a while to load in. Now these list (which are stored in array), are cached the first time the user opens the page, making it much quicker for the the dropdowns to load the next time they reload the page. Testing for this is done, by seeing how much quicker the tests run.

closes #278 
---
__Licensing Certification__


braughtg:
> This error means that I do not have permission to access /user/logout, but I am not sure why. I will continue to look into this, but any clues would be helpful!

Might it work to just cy.visit the localhost/user/logout URL instead of adding the new cypress command?

IrisSC:
Dear @braughtg,

Thank you for the advise, unfortunately I think it as more to do with going to localhost/user/logout then it does the command I use to get there. I tried using cy.visit('/user/logout'), but I am getting the same error. Any thoughts as to why this might be?

Sincerely,
Iris

braughtg:
I've run and tested this manually and it looks good.  Please do the following:
- confirm that this PR contains all of the tests that you would want to have if they were working.
- open a new issue for resolving the problems with the tests. Link that issue to this PR and the Original Issue #278. 
- convert this to a full PR.  I will then merge it.

PR #307, opened 2021-10-30
New date change report timing

### LINKED ISSUE #274: Seeding Report Page - Fix timing of report disappearing when new dates selected
The seeding report page hides everything but the date selection section when new dates are being chosen. Currently, the rest of the page disappears once the input has been changed and then clicked off of, instead of right after the initial click.
datechange

There are two possible solutions:
* Change the page to hide the report right after the user does an initial click on the date inputs
* Add a "Change Date Range" button that, when clicked, hides the rest of the page and allows the user to input new dates.

Batese2001:
__Pull Request Description__

This should resolve #274.

It should:
- change the timing of the report disappearing to when the input is clicked 

---
__Licensing Certification__

### DIFF
--- a/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
+++ b/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
@@ -29,7 +29,7 @@
                 <br>
             </div>
 
-            <div class="loader center-block" v-if="stillLoading"></div>
+            <div data-cy="loader" class="loader center-block" v-if="stillLoading"></div>
 
             <div v-if="seedingLogs.length != 0">
                 <br>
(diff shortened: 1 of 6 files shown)


braughtg:
When I test this and click on a date the report does not seem to hide and the generate report button does not reappear?  Perhaps I am doing something incorrect?  

Four points:
- Is changing the components the correct way to make this work? It seem like having the component emit a hide event is putting page specific behavior into the component.  If the component must emit an event to implement this feature then pick a more general name for the event so that it can be used for other purposes in other pages.
- It does not seem to be working for me.  When I generate a report, then click on the date picker nothing happens - the report remains visible.  Even if I change the date there is no way to generate a new report.  What is the desired behavior?
- When making a PR please be sure to remove all console.log statements and any commented out code.
- If the components must be changed, then the tests for the components should also be updated to test the new functionality (e.g. that the event is emitted at the proper time.)

Batese2001:
I have tested this now, and it does not appear that components pass their events to grandparents. I will add testing to make sure it is working, though I am not sure why it wasn't working for you, as I did seem to have committed the right files on my end. I will look into it

Batese2001:
This works on my end and passes all of my tests so it should also work on your end

PR #308, opened 2021-11-02
Eventbus spike to send event to component

Batese2001:
__Pull Request Description__

This is a collection of some unsuccessful testing to utilize an event bus to facilitate bi-directional communication between components, but this has been unsuccessful so far

---
__Licensing Certification__


Batese2001:
This is one of the websites that I used to look into event busses, but the project seems to be structured in a different way to FarmData, which may have led to some of my difficulties
https://www.digitalocean.com/community/tutorials/vuejs-global-event-bus

Batese2001:
This is the website of Vuex which I have seen several forum posts recommend using instead of an event bus, though not because others were not able to get it working
 https://vuex.vuejs.org/

braughtg:
We decided to pursue an alternative approach of creating a component to export reports.  Thus, the need to send events is less pressing.  I will close this PR, but it will be here if the need arises again.

Issue #310, opened 2021-11-10
Report Disappearing Page Testing


Batese2001:
Pull request #307 has tests in components, but does not have testing in the page. These tests should be written and should make sure that the report disappears at appropriate times.

PR #311, opened 2021-11-10
Disappear page tests

Batese2001:
__Pull Request Description__

This should add the tests for the report disappearing in the page.
Solves #310 

---
__Licensing Certification__


braughtg:
In looking at the additions to the test code, I think it has a lot of what is needed.  But I think it could test the desired behavior more directly.  

What do you think of having this test:
- generate a report and confirm that it appears
- click on one of the date elements
- confirm that the report disappears and the generate report button appears.

Then also having a second test that checks that the same behavior occurs for the other date element.

braughtg:
These changes look good.  Just remove the .only on the context and it will be ready to merge.

Issue #290, opened 2021-10-06
Seeding Report: Export 


braughtg:
There should be an "Export Report" button on the report page that will export the full report. The functionality should be:

- Button is enabled only after the report is fully generated, including the summary tables.
- Data should be exported in CSV format.
- User should be able to specify filename and location for the export (with appropriate defaults).

Batese2001:
How should I test this? Does cypress provide a way of checking downloaded files?

braughtg:
> How should I test this? Does cypress provide a way of checking downloaded files?

Good question.  Cypress has a way to read files but it looks like it can only read them within the project directory.  Because we are running FD2 in a docker instance and the project directories are mounted from the host machine it may be possible to do the download to a directory in the project (e.g. the same one with the spec in it).  Then read the file and check some of its contents.  Give it a try, but don't spend too long on this without getting back in touch to check in.

braughtg:
We should think about where and how to implement this download functionality.   It will be needed on all of the report pages so it would be great if we could implement it in an easily reusable way.  What are your thoughts on how that might be possible?

Batese2001:
Hmm, I am able to get the report table read pretty easily but the summary tables are giving me some trouble

Batese2001:
> > How should I test this? Does cypress provide a way of checking downloaded files?
> 
> Good question. Cypress has a way to read files but it looks like it can only read them within the project directory. Because we are running FD2 in a docker instance and the project directories are mounted from the host machine it may be possible to do the download to a directory in the project (e.g. the same one with the spec in it). Then read the file and check some of its contents. Give it a try, but don't spend too long on this without getting back in touch to check in.

I have been reading up on this and I can't seem to find a way to change a download file path just in javascript. From what I can tell, the browser just goes to the default downloads no matter what due to security concerns which makes sense. I wonder if there is a way to open up the file explorer to let the user pick the location manually like one might on other platforms

braughtg:
> I have been reading up on this and I can't seem to find a way to change a download file path just in javascript.

It looks like in your draft PR #297 that you say the user can specify the location and name of the download. I'm assuming that happens via dialog boxes.  Is it not possible to control those dialog boxes from the cypress tests to change the path?  

My inclination here is to not spend too much time on this at this point.  The larger issue will be how to generalize this functionality so that it is easy to incorporate it into the other reports that will be being generated.  My sense (as summarized above) would be to see if we can move the main table export function to the custom table component.

[...]

PR #297, opened 2021-10-12
Export button

Batese2001:
__Pull Request Description__

Closes #290 

It should:
-Add Export Button after report and summary tables are generated
-Export a CSV file to a location specified by the user and with a custom name
-Default sensibly

---
__Licensing Certification__

FarmData2 is a [Free Cultural Work](https://freedomdefined.org/Definition) and all accepted contributions are licensed as described in the LICENSE.md file. This requires that the contributor holds the rights to do so. By submitting this pull request __I certify that I satisfy the terms of the [Developer Certificate of Origin](https://developercertificate.org/)__ for its contents.

PR #313, opened 2021-11-12
Download Button Component

IrisSC:
__Pull Request Description__

Created a Download Button Component that when clicked will download information given to it through the page and turn it into a csv that is downloaded into the users download folder. This pull request includes the code for the component as well as the cypress tests for the component. 

closes: #290 

References Pull Request: #297 

---
__Licensing Certification__


IrisSC:
Hello,

I am having some issues with the tests for the export button component, specifically in the beforeEach section of the tests. I am trying to load the props into the component (specifically the headers and rows), but I am getting this error:
image
I am not sure what this error means or how to fix it. One thing that I have noticed is that when the array data in the row objects is empty the error goes away. I am assume that the error has something to do with handling that information, but I am not sure what. 
image
I am using the same sample data that is used in the custom table component tests, since the CustomTableComponent takes in the same data for the rows and headers props as the ExportButtonComponent does. The tests for the CustomTableComponent work just fine, so I am not sure why the ExportButtonComponent is having this problem.

Any help would be wonderful!

Sincerely,
Iris

IrisSC:
I am going to work on test that do not need anything in the data array of the rows, so on new commits there will not have the error, but the issue still has not been fixed.

farmdata2_modules/cypress/downloads/tableDownload.csv
```diff
@@ -0,0 +1,3 @@
+cool,works?,hello,
```
braughtg, line 1: Looks like there is an extra comma at the end of the line.  You'll want to be sure that isn't there on the headers or the data rows in the final verson.

farmdata2_modules/fd2_tabs/resources/ExportButtonComponent.js
```diff
@@ -0,0 +1,86 @@
...
+            })
+            return window.URL.createObjectURL(data);
+        },
+    },
+    computed:{
+        isVisible() {
+            let updatedVis = []
+            if (this.visibleColumns == null) {
+                for (i = 0; i < this.headers.length; i++) {
+                    updatedVis.push(true);
```
braughtg, line 70: It seems like here if there are 3 headers then automatically the first 3 data columns will be displayed instead of the 3 that correspond to the locations of the headers?  For example if headers has length 3, then updatedVis will be [true, true, true].  If the downloadRows has 5 columns then just the first 3 will be displayed.  

But maybe I'm missing something.  Perhaps the downloadRows are also already pruned in the Prop?  If that is the case, then maybe this part isn't necessary at all?

braughtg:
> Any help would be wonderful!

I don't know for sure what the issue is here.  But I think having a look at the "stack trace" of the error in Cypress may provide some more information.  If we can track it down to the line where the error occurred that might get you going in the right direction.

PR #314, opened 2021-11-16
Seeding Log Creation Confirmation

### LINKED ISSUE #285: Seeding Log Creation Confirmation
When a new seeding log is submitted via the "Seeding Report" sub-tab of the "Field Kit", a confirm dialog comes up.  This makes it apparent to the user that the log was created. We should think about the default behavior and button options on this dialog.  For example:
* Should the dialog be confirm - Create seeding log? [OK | Cancel], then the user could cancel the creation if the submit button was clicked/tapped by accident.
* What should be the default behavior after OK or Cancel is clicked?
  * For Cancel seems it should return to the seeding log form with the current values.
  * For OK, should it go back to a blank seeing input form so another entry could be created quickly?  What is the common workflow? Do they typically create multiple seeding logs in a row?

Would be a good conversation to have with Matt.   

Whatever is ultimately implemented here should also have Cypress tests that check the behavior.

Batese2001:
__Pull Request Description__

Resolves #285 
It will:
- create logs on a dialogue box confirmation
- alert the user that this was successful with some kind of pop up
- reset all values to their initial state on confirmation
- leave them as they were on rejection
- include tests for these

---
__Licensing Certification__

### DIFF
--- a/farmdata2_modules/fd2_tabs/fd2_field_kit/seedingInput.html
+++ b/farmdata2_modules/fd2_tabs/fd2_field_kit/seedingInput.html
@@ -9,7 +9,7 @@
                 <br>
                 <dropdown-with-all data-cy="crop-selection" :dropdown-list='cropArray' @selection-changed='setNewCrop'>Crop:<label style='color: #da4f3f'>*</label></dropdown-with-all>
                 <br>
-                <dropdown-with-all data-cy="area-selection" :dropdown-list='areaArray' @selection-changed='setNewArea'>Area:<label style='color: #da4f3f'>*</label></dropdown-with-all>
+                <dropdown-with-all data-cy="area-selection" :dropdown-list='areaFilter' @selection-changed='setNewArea'>Area:<label style='color: #da4f3f'>*</label></dropdown-with-all>
                 <br>
                 <label>Comments: </label><textarea data-cy="comments" v-model='comments'></textarea>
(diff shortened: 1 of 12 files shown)


Batese2001:
This last commit does not reset the dropdowns. Do we want the dropdowns to be reset as well?

farmdata2_modules/fd2_tabs/fd2_field_kit/seedingInput.html
```diff
@@ -83,8 +101,36 @@ <h1 style='text-align:center;'>Seeding Input Log</h1>
...
+                stopAlert(){
+                    this.alertVal = false
+                },
+                clearForm(){
+                    this.cropArray.push('')
+                    this.areaFilter.push('')
+                    console.log(this.cropArray)
+                    console.log(this.areaFilter)
+                    this.selectedDate = dayjs().format('YYYY-MM-DD').toString()
+                    this.selectedCrop = "ARUGULA"
```
braughtg, line 116: Should the selectedCrop be set to '' here in the same way the selectedArea is below?

braughtg:
WIth #350 merged the issues with this PR not resetting dropdown elements should be fixed.  There will still be issues with resetting of date elements that will require #351 and #352 to be resolved.

braughtg:
* Superseded by #285.

Issue #319, opened 2021-12-02
DropdownWithAllComponent Selection Should Change when Prop Changes


Batese2001:
When the defaultInput prop in the DropdownWithAllComponent is changed outside of the component, the selected value remains the same. When the prop is changed in the containing page, the selected value should change.

PR #320, opened 2021-12-02
Custom Dropdown Component props changes selection

Batese2001:
__Pull Request Description__

This should make the defaultInput and defaultDate props change the selection even when changed outside of the component

Solves #319 

---
__Licensing Certification__

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -16,13 +16,15 @@ Connect with the [FarmData2 community on Zulip](https://farmdata2.zulipchat.com/
 
 If you are unfamiliar with [Zulip](https://zulip.com/) it is a group chat application that blends the benefits of threaded discussions with live chat. Zulip is relatively easy to use once you understand the key ideas of [streams and topics](https://zulip.com/help/about-streams-and-topics).
 
+## Developer Installation ##
+
+Having a running version of FarmData2 is a prerequisite for many of the forms of participation described below.  The [Install Directions] give step by step instructions for getting a developer installation of FarmData2 up and running.
+
+[Install Directions]: INSTALL.md
+
 ## Participation ##
 
 There are many ways to participate in FarmData2. Some of them are listed below.  
-
-Having a running version of FarmData2 is a prerequisite for many of the forms of participation.  The [Install Directions] give step by step instructions for getting FarmData2 up and running.
-
-[Install Directions]: INSTALL.md
 
 #### Bug Reports ####
 
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,4 +1,4 @@
-## Installing FarmData2 ##
+# Installing FarmData2 #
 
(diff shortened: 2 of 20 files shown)


Batese2001:
Currently, the default input overrides what is chosen in the dropdown. Is it a good idea to create a new prop or method within the component that governs clearing the values to the initial?

braughtg:
> Currently, the default input overrides what is chosen in the dropdown, which, while allowing us to change the selection from the outside, does not let us do the same from the inside.

Would using v-bind to connect the prop to the dropdown instead of v-model solve this problem?

braughtg:
You will also want to add some component tests for the change to the associated spec files in FarmData2/farmdata2_modules/fd2_tabs/resources/

Then run them with the component testing framework.  See the documentation for component testing in: https://github.com/DickinsonCollege/FarmData2/tree/main/farmdata2_modules/fd2_tabs/fd2_example

Batese2001:
> > Currently, the default input overrides what is chosen in the dropdown, which, while allowing us to change the selection from the outside, does not let us do the same from the inside.
> 
> Would using v-bind to connect the prop to the dropdown instead of v-model solve this problem?

I have tried using the v-bind and v-model, but it seems that either they are unchanged by the outside value after it is initialized or they are constantly overwritten by the outside value and thus take several clicks to actually get what you want in the dropdown. I tested bringing out the template code into the seeding input form and everything worked fine, so it definitely seems to be due to how Vue components work, is my thinking

braughtg:
> it definitely seems to be due to how Vue components work, is my thinking

Can you add some Cypress component tests that illustrate the issue?  Try to have he test try to set the selected element from outside of the component and have the test fail if the component does not behave the way we want it to from the outside.  Then maybe another were the selected element is set by clicking - though I think that may already be one of the existing tests.

braughtg:
Another thing that I just came across suggests that you may need to pass the prop something like:

`:default="'value'"`

Note the " and then the ' in the value.  This seems to suggest that the value you set for selected inside the component must be a string. For that to be the case the prop going in must be a string that contains a string.  Not sure if that will work or not but its a pretty easy thing to try.

braughtg:
If we need more information, this series of examples seems like it would be helpful to ensure that we are setting up the component in the correct way:

https://designhammer.com/blog/reusable-vuejs-components-part-2-basic-drop-down-and-v-model

braughtg:
> > > Currently, the default input overrides what is chosen in the dropdown, which, while allowing us to change the selection from the outside, does not let us do the same from the inside.
> > 
> > 
> > Would using v-bind to connect the prop to the dropdown instead of v-model solve this problem?
> 
> I have tried using the v-bind and v-model

Based on some further reading, my suggestion to use v-model here was incorrect.  v-bind is the correct way to do this.

Using v-model would cause the prop value outside of the component to be changed when the selected item is changed.  The right way to handle that is via the emitted event - which we do correctly already.

So again v-bind was the proper way to do this here.  Sorry for the bad directions.

Batese2001:
The tests added to the component should run without error when the issue is solved, but currently do not run correctly. When the test runs correctly, the issue can be closed. The date selection component must also be fixed

braughtg:
Superseded by #350

PR #478, opened 2022-06-08
Seeding Input Form UI Design

### LINKED ISSUE #301: Seeding Input Form UI Design
Consider moving the Comment's field for the Seeding Input form on the Field Kit tab to the bottom of the form, in its own box.  Add a red heading "Comments" and have it just contain the comment box without its own label.  Comments will be rarely used. So moving this to the bottom will prevent having to scroll over that box while entering data for a planting.  It will still need to be scrolled over at the end to get to the submit button, but that will not be as disruptive to the workflow.

JingyuMarcelLee:
__Pull Request Description__

Fixes #301 Seeding Input Form UI Design. UI became more concise and transparent. Some additional issues, such as the sizing of the date and dropdown components may need to be addressed in the future through changes in the Vue components. The PR says that the component file for the dropdown was altered, but I reverted the change. 

Co-authored by @FutzMonitor 

---
__Licensing Certification__


braughtg:
This looks really good.  Now having seen it live there is one other change that I think would be good.

I think making the bed/row drop down in the Direct Seeding view a drop down with Row Feed / Bed Feet (removing the "Feet" text label) will make this more clear to the user and a give the UI a little cleaner look.

PR #496, opened 2022-06-22
Creation of new Vue Comp: RegexInputComponent

### LINKED ISSUE #333: Seeding Input: Allows invalid input
**Description:** When creating a new Seeding Input Log, it is possible to enter values that are invalid, such as negative numbers and decimal values that do not make sense in certain categories (e.g. Workers: 9.4).

**Directions to Replicate the Bug**
1. Open the FarmData2 Application
2. Click on the *Seeding Input* button in the *FieldKit* Tab
3. Select any *Date of Seeding*, *Crop*, *Area*, and choose either *Tray Seeding* or *Direct Seeding*
4. Enter invalid values, such as negative numbers or decimal values
5. Click *Submit*

**Expected Results:** The application should not let the user submit these values and should give the user an error message that notifies them of the appropriate values that they should input.

**Observed Results:** The application allows the user to submit these invalid values and stores this Seeding Input Log with the rest of the data.
image

FutzMonitor:
__Pull Request Description__

Resolves #333. The issue originally would've seen some modifications to the parent page to simply restrict/make it impossible for users to input invalid data into the Seeding Input form. However, as development continued it became more apparent that this problem might appear else where in FarmData2, so we encapsulated its solution into a Vue component that can be implemented wherever need be. 

Simply put, the new _RegexInputComponent_ allows developers to specify a regex that would be compared to the value inputted by users inside an input html element. If the user input is valid, then the component returns "true" and allows the value to be passed to whatever it needs to be passed to depending on parent page implementation. If the user input is invalid, then the component returns "false" and that is also handled depending on the parent page's implementation. 

---
__Licensing Certification__


farmdata2_modules/fd2_tabs/fd2_field_kit/fd2_field_kit.info
```diff
@@ -3,4 +3,5 @@ description = A collection of forms for fast and accurate data entry in the fiel
 core = 7.x
 package = FarmData2
 
-scripts[] = '../resources/FarmData2.js'
\ No newline at end of file
+scripts[] = '../resources/FarmData2.js'
+scripts[] = '../resources/RegexInputComponent.js'
```
braughtg, line 7: Let's remove all changes to the FieldKit for this PR.  We'll update the FieldKit to include this component in the PR for #285.

farmdata2_modules/fd2_tabs/resources/RegexInputComponent.js
```diff
@@ -0,0 +1,152 @@
...
+ * @vue-prop {String} [setColor] - Defines the background color of the input element.
+ * @vue-prop {String} [setHeight] - Defines the height of the input element.
+ * @vue-prop {String} [setWidth] - Defines the width of the input element.
+ * @vue-prop {String} [setMin] - Defines the min of the input element.
+ * @vue-prop {String} [setMax] - Defines the max of the input element.
+ * @vue-prop {String} [setStep] - Defines the step increments of the input element.
+ * @vue-prop {String} [setTitle] - Defines the title for a simple hover tip for the input element.
+ * @vue-prop {Boolean} [isDisabled] - Makes the element read-only.
+ * 
+ * @vue-event {Boolean} is-enabled - Emits boolean (isMatch) for the parent page to make use of the results of the validation.
```
braughtg, line 26: It looks like you have two events emitted in the code.  This documentation should match what the code is doing.

farmdata2_modules/fd2_tabs/resources/RegexInputComponent.js
```diff
@@ -0,0 +1,152 @@
...
+ * </tbody>
+ * </table>
+ * 
+ * @vue-prop {RegExp} [regExp] - The regular expression used to validate user inputs.
+ * @vue-prop {String} [defaultVal] - Used to clear the page once a log has been submitted.
+ * @vue-prop {String} [setType] - Defines the type of input box.
+ * @vue-prop {String} [setColor] - Defines the background color of the input element.
+ * @vue-prop {String} [setHeight] - Defines the height of the input element.
+ * @vue-prop {String} [setWidth] - Defines the width of the input element.
+ * @vue-prop {String} [setMin] - Defines the min of the input element.
```
braughtg, line 20: Min, max, step props should indicate that they are ignored if not applicable to the type of the input element.

farmdata2_modules/fd2_tabs/resources/RegexInputComponent.js
```diff
@@ -0,0 +1,152 @@
...
+              height: this.setHeight,
+              width: this.setWidth,
+            },
+          }
+        },
+        methods: {
+          // Handles validation of the input value
+          blurEventHandler: function (e) {
+            const inputVal = e.target.value;
+            console.log(inputVal);
```
braughtg, line 98: Be sure to remove the console.log's from your production code.

farmdata2_modules/fd2_tabs/resources/RegexInputComponent.spec.comp.js
```diff
@@ -0,0 +1,228 @@
...
+                    .blur()
+                    .then(() => {
+                        expect(spy).to.be.called
+                    })
+    })
+})
+
+    context('testing disabled option', () => {
+        let comp;
+        beforeEach(() => {
```
braughtg, line 43: Should this be in before each?  If you shallow mount before each and then mount in the "it" below isn't the component mounted twice?  Once shallow and once normal?  Maybe you don't need the before each here and just mount or shallow mount in the it?  Would that work?

farmdata2_modules/fd2_tabs/resources/RegexInputComponent.spec.comp.js
```diff
@@ -0,0 +1,228 @@
...
+                        setMax: null,
+                        setStep: null,
+                        setTitle: null,
+                        disabled: null
+
+                    },
+                })
+            })
+
+            it('change regExp prop', () => {
```
braughtg, line 94: Do all of these prop changes work?  I don't see watch statements for all of them in the component code.  Isn't it necessary to have a watch in the component for the change in prop to be detected in the component?  If these tests pass it seems not, maybe we need to investigate this some more?

braughtg (changes requested):
See the comments in the code above.

farmdata2_modules/fd2_tabs/resources/RegexInputComponent.spec.comp.js
FutzMonitor: When you remove the _before()_ from the context, the following test stops working: 

image

I'm guessing when it isn't shallow mounted before each test, the test that changes the data value via the prop stops working because you need a shallow prop in order to change prop values which will trigger a watch that will change the data value.

PR #497, opened 2022-06-22
Issue #357 CustomTableComponent Prop Change reponse

### LINKED ISSUE #357: CustomTableComponent Should Respond to Prop Changes
The CustomTableComponent (in farmdata2_modules/fd2_tabs/resources) should be updated so that changing the visibleColumns, canEdit, canDelete, headers props cause table to update as appropriate.
- Update associated component tests.
- Expand the UI example page and tests to demonstrate some of the capabilities (e.g. showing and hiding a column)
- Update the documentation in the README.md in resources.

See:
- #350 for how the DropDownWithAllComponent was changed to respond to prop changes.

JingyuMarcelLee:
__Pull Request Description__

Fix #357. Added watch for visible columns and tested prop Changes. 

---
__Licensing Certification__


JingyuMarcelLee:
Ready to be reviewed and merged

braughtg:
Looks like your tests check that the watch is invoked and the UI change occurs when the entire array (i.e. the reference in the prop) is changed.  Can you create a test that also checks that everything works if the whole prop array is not changed but one of its elements is changed?  I imagine a common use case will be for the page to hold an array of True/False values and then one of those values will be changed but the prop will still refer to the same array. 

I think you'll need something like this:
* https://michaelnthiessen.com/how-to-watch-nested-data-vue/

Issue #501, opened 2022-06-22
UI Examples Tab: Add example of the RegExInputComponent


braughtg:
Add an example of using the RegExInputComonent to the UI tab in the FD2 Examples tab. The example should:
* include a regular expression that only allows some format of input (e.g. positive integers). 
* display the events that are emitted in the page (similar to how other components display values from emitted events)
* include a button that disables/enables the field using its prop.
* include a button that sets the value of the field using its prop.

Cypress tests should be added to UI.spec.js to test the added functionality.

FutzMonitor:
I will be tackling this issue.

PR #504, opened 2022-06-27
Updating the UI Page

### LINKED ISSUE #484: UI Examples Tab: Add buttons to set dates for Date components using prop change.
The DateSelectionComponent and DateRangeSelectionComponent have been updated to respond to changes in their prop values (#351, #352).  Buttons should be added beside each of these components to change their values using a change to their prop.  This will be similar to what is done with the "Choose Kale" button for the Dropdown.  It will demonstrate that the value of the component can be changed by code in the page that contains it.  Corresponding tests for these new buttons should be added to the UI.spec.js file as well.

FutzMonitor:
__Pull Request Description__

Resolves #484 & #501. Adds the RegexInputComponent to the UI page for user to acquaint themselves with the component. Added Cypress e2e tests for this new addition to the UI sub-tab. 

Additionally, all components in the UI sub-tab have had a *disabled* button added to their demonstration that shows what happens when you disabled the components. The appropriate Cypress e2e tests have been added as well. 

---
__Licensing Certification__


farmdata2_modules/fd2_tabs/fd2_example/ui.html
```diff
@@ -176,11 +213,56 @@
...
+        setNewRegexMatch(newMatch){
+            this.regexMatchVal = newMatch
+        },
+        setNewRegexVal(newVal){
+            this.regexInputVal = newVal 
+        },
+        resetRegexVal(){
+            this.regexInputVal = '0' 
+        },
+        pickRegExp(newOption) {
```
braughtg, line 251: I would suggest using a map to hold the regex and then pass the key to this function from the @click event.   That way there is no if statement in this function and the calls to it from the two buttons are consistent.

farmdata2_modules/fd2_tabs/fd2_example/ui.html
```diff
@@ -176,11 +213,56 @@
...
     },
     methods: {
+        checkDisabled(key) {
+            if(this.disabledMap[key]) {
+                return "Enable"
+            } else {
+                return "Disable"
+            }
+        },
+        makeDisabled(key) {
```
braughtg, line 236: This function name doesn't seem to accurately describe the function's purpose.  Maybe "toggleDisabled"?

farmdata2_modules/fd2_tabs/fd2_example/ui.html
```diff
@@ -176,11 +213,56 @@
...
+        regexMatchVal: false,
+        disabledMap: {
+            "DropdownWithAllComponent": false,
+            "RegexInputComponent": false,
+            "DateSelectionComponent": false,
+            "DateRangeSelectionComponent": false,
+        }
     },
     methods: {
+        checkDisabled(key) {
```
braughtg, line 229: This function name does not seem to describe the function's purpose.  It seems to return the label for the disable/enable button.  Is there a better name?  Alternatively, I think it would be more "Vue-like" to do this with 2 buttons and a v-if.

farmdata2_modules/fd2_tabs/fd2_example/ui.html
JingyuMarcelLee: Changed function name as suggested.

PR #532, opened 2022-07-12
Transplanting Input Form

### LINKED ISSUE #526: Transplanting Input Form To Log Transplanting of Crops
## Current Design ## 
The Active Planting Report (#416) is going to need a way for workers to be able to log the amount of crops they transplanted. Currently, there is no way of doing that since the Fieldkit module only has a Seeding Input Form sub-tab but no input form for transplantings.

| image | 
|:--:| 
| *current view of the Fieldkit module* |

## Desired Design ## 
1. The [fd2_field_kit.module](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/fd2_field_kit/fd2_field_kit.module) will need to be modified to support a new sub-tab within the Fieldkit module. 
2. A new HTML file will need to be added into the fd2_field_kit directory (named something like transplantingInput.html) which will hold the HTML, Vue, and JavaScript for the page. 
3. A new Cypress e2e test file will need to be added into the fd2_field_kit directory to test the page's functionality.

| image | 
|:--:| 
| *what a potential transplanting input form could look like* |
- This proof of concept was provided by @katiele47 and her teammates. 
- her and her team's code can be found [here](https://github.com/hantrinh13/FarmData2)

FutzMonitor:
__Pull Request Description__

Closes #526. This PR hopes to add a Transplanting Input form sub-tab to the Field Kit module. When you visit the page, a table with planting assets exclusively in tray seeded areas will appear and clicking a button will bring up the transplanting input form pre-populated with the information of that asset's tray seeding log. From there the page behaves the same as the seeding report minus user editing. The user has to submit a transplanting log solely from the pre-populated values done automatically between the table and the input form (save for some specifics like labor, comments, and the amount planted). This is to address the possibility of users creating multiple logs for the same crop in the same area. 

**This PR is not ready to be merged into upstream/main**
Here are a list of things that need to be accomplished before this PR is ready to be reviewed: 
- [ ] Complete the functionality of the table that displays the planting assets
- [ ] Complete the Transplanting Input form's functionality (API calls, necessary functions, etc.)
- [ ] Complete the communication between these two features
- [ ] Complete the testing file for this page
- [ ] Cleanliness: spaces, indentation, and consistency
- [ ] Change the first commit and co-author original authors of the base code

---
__Licensing Certification__


braughtg:
@FutzMonitor  said "Currently it takes a good five seconds for the assets to load in. I believe it'll be worthwhile to cache them, but I'm aware that I'm going to have to implement something to remove it from the local cache if that asset is no longer in the list. Is that possible?"

Caching is one idea, and would be useful once records have been fetched.  We might also allow the user to set the crop and/or area before fetching the list.  That would reduce the number of records to be fetched and should speed things up.  So the form might include a button to fetch the assets. It would then do that based on the values in the dropdowns indicating the selected crop and/or area.  Thoughts?

FutzMonitor:
> @FutzMonitor said "Currently it takes a good five seconds for the assets to load in. I believe it'll be worthwhile to cache them, but I'm aware that I'm going to have to implement something to remove it from the local cache if that asset is no longer in the list. Is that possible?"
> 
> Caching is one idea, and would be useful once records have been fetched. We might also allow the user to set the crop and/or area before fetching the list. That would reduce the number of records to be fetched and should speed things up. So the form might include a button to fetch the assets. It would then do that based on the values in the dropdowns indicating the selected crop and/or area. Thoughts?

Would this be similar to the Seeding Report using dates to generate the report table?

braughtg:
> Would this be similar to the Seeding Report using dates to generate the report table?

Yes it would be similar to that if this seems a reasonable way to go.  We could try it and experiment with it and see how it works and feels as a user.

FutzMonitor:
> > Would this be similar to the Seeding Report using dates to generate the report table?
> 
> Yes it would be similar to that if this seems a reasonable way to go. We could try it and experiment with it and see how it works and feels as a user.

The original plan that I had in my mind is that I would have a sort of generate report bar, similar to the Seeding Report, with the crop and area dropdown. A generate button would light up when a selection has been made, and it would generate a table of assets fitting the criteria specified by the user. When the button was clicked, it would pass the selections to a method that would make an API request for all planting assets that matches that criteria. The problem is that is seems that you can't go into the arrays containing the information for the crop and area using an API request: 

image
image

The other implementation would have to just generate a report table when the user visits the page of all the planting assets that are currently in the ground. This was previously going to be done using the `done` data to determine if the planting asset has been harvested or not. Unfortunately, that cannot be done because that data doesn't exist for the planting asset.  If it did exist, we could call all those assets and cache them and simply update that cache by removing assets that have been marked as done. However, without it, we basically have to cache all assets which will grow more and more inefficient and cost space as the assets grow during the lifetime of a farm.

braughtg:
With the reorganization of the code this PR will ultimately need to be reworked to place the input form in the proper location.

PR #540, opened 2022-07-18
Issue #536 Labor data summary fix

### LINKED ISSUE #536: Seeding Report Summary Labor Data Calculation after Hidden/Optional Labor Configuration
## Description of the Issue 

After allowing users to make configuration changes to hide the display of labor-related fields on both seeding input form and the seeding report (Issue #520 #519 covered by #515) , a new problem has arisen. When the logs without labor data created with "hidden" or "optional" labor configuration are displayed in the seeding report, the report automatically converts the N/A values to 0, using this in the calculation of the data summary on the summer table at the bottom of the page.

## Steps to Reproduce

1. Go to the configuration page and change the setting to Hidden or Optional. Save the change.

2. Go to the Seeding Input log and create a log without the labor data (leaving it blank)

3. Go back to the configuration page and set the labor back to required (so you can see labor-related data on the report)

4. Visit Seeding report and enter today's date. Generate the report. The labor data shows 0 instead of N/A. The summary table uses this value to make calculations for total and average hours. 

5. The actual log has N/A for number of workers and hours. 

## Suggested Solution
Fix the computed properties in the seeding report so that N/A values are excluded in its calculation. Before that, we need to check if the JSON for the seeding log actually saves an empty value as 0 by default. If this is the case, this has to be addressed in the seeding input so that empty values do not get saved as 0.

JingyuMarcelLee:
__Pull Request Description__

Fixes #536 Completed fixing the bug where the summary table displayed incorrect labor data when seeding input had no labor data. 
---
__Licensing Certification__

### DIFF
--- a/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
+++ b/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
@@ -506,11 +506,8 @@
                     let tableData = this.tableRows
                     let sumRowFeet = 0
                     tableData.map(h => {
-                        if(parseFloat(h.data[10]) == 0 && parseFloat(h.data[11]) == 0) {
-                            sumRowFeet = sumRowFeet
-                        } else {
-
-                            sumRowFeet = sumRowFeet + parseFloat(h.data[5])
+                        if(parseFloat(h.data[10]) != 0 && parseFloat(h.data[11]) != 0) {
+                            sumRowFeet = sumRowFeet + parseFloat(h.data[5])        
                         }
                     })
                     return (Math.round(sumRowFeet*100))/100
@@ -533,9 +530,7 @@
                     let tableData = this.tableRows
                     let sumBedFeet = 0
                     tableData.map(h => {
-                        if(parseFloat(h.data[10]) == 0 && parseFloat(h.data[11]) == 0) {
-                            sumBedFeet = sumBedFeet
-                        } else {
+                        if(parseFloat(h.data[10]) != 0 && parseFloat(h.data[11]) != 0) {
                             sumBedFeet = sumBedFeet + parseFloat(h.data[5])
                         }
                     })
@@ -583,9 +578,7 @@
                     let tableData = this.tableRows
                     let totalTraySeeds = 0
                     tableData.map(h => {
-                        if(parseFloat(h.data[10]) == 0 && parseFloat(h.data[11]) == 0) {
-                            totalTraySeeds = totalTraySeeds
-                        } else {
+                        if(parseFloat(h.data[10]) != 0 && parseFloat(h.data[11]) != 0) {
                             totalTraySeeds = totalTraySeeds + parseFloat(h.data[7])
                         }
                     })


farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
```diff
@@ -517,35 +549,57 @@ <h1 class="text-center">Seeding Report</h1>
                     let filteredTableData = tableData.filter((x) => x.data[3] === 'Direct')
                     let sumHours = 0
                     filteredTableData.map(h => {
-                        sumHours = sumHours + h.data[10]//*h.data[11])
+                        sumHours = sumHours + (h.data[10]*parseFloat(h.data[11]))
```
JingyuMarcelLee, line 552: I don't know why this was commented out originally, but it shouldn't be commented out.

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js
```diff
@@ -526,13 +530,13 @@ describe('Testing for the seeding report page', () => {
...
 
             cy.get('[data-cy=direct-total-hours]')
-            .should('have.text', '7')
+            .should('have.text', '2.1')
 
             cy.get('[data-cy=direct-total-rowft-hour]')
-            .should('have.text', '3739.57')
+            .should('have.text', '4191.9')
 
             cy.get('[data-cy=direct-total-bedfr-hour]')
```
JingyuMarcelLee, line 538: Because I fixed the commented-out portion earlier, the values on the summary table changed (now it is correct). I had to modify the test to make it pass with the correct value (the test was incorrect before).

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
```diff
@@ -494,7 +495,23 @@ <h1 class="text-center">Seeding Report</h1>
...
+                    return (Math.round(sumRowFeet*100))/100
+                },
+                /**
+                 * returns the total row feet for Direct Seeding
+                 */ 
+                totalRowFeetNoLabor(){
+                    let tableData = this.tableRows
+                    let sumRowFeet = 0
+                    tableData.map(h => {
+                        if(parseFloat(h.data[10]) == 0 && parseFloat(h.data[11]) == 0) {
```
braughtg, line 509: You can invert the if condition here and just use an if instead of an if/else.  The if part will then add to sumRowFeed and the statement in the if: sumRowFeet = sumRowFeet that does not do anything will be removed.

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
```diff
@@ -505,7 +522,22 @@ <h1 class="text-center">Seeding Report</h1>
...
+                },
+                /**
+                 * returns the total bed feet for Direct Seeding excluding the logs without labor data
+                 */
+                totalBedFeetNoLabor(){
+                    let tableData = this.tableRows
+                    let sumBedFeet = 0
+                    tableData.map(h => {
+                        if(parseFloat(h.data[10]) == 0 && parseFloat(h.data[11]) == 0) {
+                            sumBedFeet = sumBedFeet
```
braughtg, line 537: See comment above about simplifying this if/else down to an if.

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
```diff
@@ -517,35 +549,57 @@ <h1 class="text-center">Seeding Report</h1>
...
                 },
+
+                /**
+                 * returns the total number of seeds planted for the tray seeding logs excluding the logs without labor data
+                 */ 
+                traySeedsPlantedNoLabor(){
+                    let tableData = this.tableRows
+                    let totalTraySeeds = 0
+                    tableData.map(h => {
+                        if(parseFloat(h.data[10]) == 0 && parseFloat(h.data[11]) == 0) {
```
braughtg, line 586: Also simplify this if/else.

PR #555, opened 2022-07-25
Removed Anonymous from the User List in Seeding Report

### LINKED ISSUE #554: 'Anonymous' User Causes a 406 Status Code in the Seeding Report
## Current Behavior ##
1. Visit the Seeding Report
2. Open up the dev tools either by right clicking and clicking "inspect" or pressing F12 on most browsers. 
3. Generate a report.
4. Click the edit button and inside of the user dropdown select the option 'Anonymous'.
5. Click save.

Your console should have printed something like the following: 

image

## Desired Behavior ##
It would actually be better to just not have 'Anonymous' load into the user list altogether. However, it seems to come along with the Drupal package. So for this fix it would be best if the 'Anonymous' user isn't loaded into the dropdown to avoid this problem. The fix would look something like this: 

before: 
image

after: 
image

FutzMonitor:
__Pull Request Description__

Closes #554 

---
__Licensing Certification__


FutzMonitor:
I'll flip it once I removed some things that accidentally came over from another branch.

braughtg:
I think we should consider removing Anonymous from the IDToUserMap and the UserToIDMap in the FarmOSAPI.js functions instead of in each individual place within FarmData2 where it might need to be removed.

FutzMonitor:
This branch is ready to be reviewed once again.

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
```diff
@@ -733,6 +733,7 @@ <h1 class="text-center">Seeding Report</h1>
                     for(i=0; i < this.userToIDMap.size; i++){
                         userNameArray.push(userKeyIterator.next().value)
                     }
+                    userNameArray.splice(0, 1)
```
braughtg, line 736: Should this be here? Looks like it will be cutting out a name?

FutzMonitor: This has now been resolved, sorry about that.

farmdata2_modules/fd2_tabs/resources/FarmOSAPI.js
```diff
@@ -95,7 +95,22 @@ function getAllPages(endpoint, arr=[]) {
  * // Note that the Map will not be available until the then() executes.
  */
 function getIDToUserMap(){
-    return getMap('/user', 'uid', 'name')
+    return new Promise((resolve, reject) => {
```
braughtg, line 98: Seems like we would still want to use getMap here and in the .then just remove the anonymous user?  That would allow us to not duplicate code from the getMap function.  Is there a reason for not doing it that way?

braughtg (changes requested):

braughtg:
Also, does this impact the tests for FarmOSAPI.js at all?

farmdata2_modules/fd2_tabs/resources/FarmOSAPI.js
FutzMonitor: We tried to do something along the lines of getMap(...).then((resp) => {resp.delete('Anonymous')}) but this kept causing weird errors like "(...).delete is not a function" or an empty user dropdown inside of the Seeding Report.

FutzMonitor:
> Also, does this impact the tests for FarmOSAPI.js at all?

I do have to update them, I forgot about that.

braughtg:
Also, there are no content changes to the documentation, so let's remove those from the PR so as not to create any merge conflicts with other work on #561  that will be changing the docs.

FutzMonitor:
> Also, there are no content changes to the documentation, so let's remove those from the PR so as not to create any merge conflicts with other work on #561 that will be changing the docs.

Those will be removed prior to merging.

farmdata2_modules/fd2_tabs/resources/FarmOSAPI.js
braughtg: I think you want to do is process the map in the function declared in the then and then resolve(...) the map.  So something like:
```
...
getMap(xxxx)
.then((map) => {
  // remove the thing from map
  resolve(map)
}
...
```

FutzMonitor: This is what I wrote: 
```
function getUserToIDMap(){
    return getMap('/user', 'name', 'uid')
    .then((map) => {
        map.delete('Anonymous')
        resolve(map)
    })
}
```
But this produced an error: 
image

braughtg: Give something like this a try:
```
function getIDToUserMap(){
    //return getMap('/user', 'uid', 'name')
    return new Promise((resolve, reject) => {
        getMap('/user', 'uid', 'name')
        .then((map) => {
            map.delete(null);
            resolve(map)
        },
        (error) => {
          reject(error);
        })
    })
}
```
It follows the same pattern you had but doesn't repeat the construction of the map, making use of the getMap function instead.

FutzMonitor: It generates the same network banner error.

braughtg: I just tried that function and the corresponding one modified for getUserToIDMap and it seems to be working fine.  Do you get that banner when the page is loaded? Or when you do something else?

FutzMonitor: On page load, let me try something to see if I can make it work.

FutzMonitor: It's working for me now after resetting the Drupal cache. Is there any chance that loading that page once caches that data?

braughtg: The script could have been cached and not updated when you changed it.  Though, I didn't have to clear the cache for my edits to take effect.  

Anyway, as long as its working now that's good.

braughtg (approved):

PR #570, opened 2022-09-17
Refactoring CustomTableComponent (and other updates)

### LINKED ISSUE #530: CustomTableComponent Refactoring
The CustomTableComponent should include the ability to specify buttons to appear in the table. These will be like the Edit/Delete buttons but will allow developers to add additional buttons that perform custom actions.  Two possible use cases include:
* Adding a "Transplant" button to a table in the Transplanting Log page that allows the user to click the button on a row to pre-populate a transplant input form for creating the transplanting log.
* A similar feature would also be useful for the Harvest Log page.

FutzMonitor:
__Pull Request Description__

Closes #530. This pull request seeks to cover numerous objectives concerning the refactoring of the CustomTableComponent. 

**Note:** This PR isn't ready to be merged yet. The following tasks must be completed prior to code review and merging: 
- [x] [CustomTableComponent.js](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/updatedCustomTable/farmdata2_modules/fd2_tabs/resources/NewCustomTableComponent.js) has to be refactored to no longer use the `headers`, `inputOptions`, and `visibleColumns` props. These changes will ripple and require modifications of the component on top of the additions of checkboxes, regex input, and custom buttons. 
- [x] Redo the documentation for both the refactored CustomTableComponent and the RegexInputComponent and regenerate the documentation in the project.
- [x] The [CustomTableComponent.spec.comp.js](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/updatedCustomTable/farmdata2_modules/fd2_tabs/resources/NewCustomTableComponent.spec.comp.js) has to be updated to repair tests that broke due to prop and function changes and new tests have to be written to test the added functionalities. 
- [x] An example of the refactored CustomTableComponent added to the [UI.html](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/updatedCustomTable/farmdata2_modules/fd2_tabs/fd2_example/ui.html) file to showcase its new capabilities alongside modifications to the [UI.spec.js](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/updatedCustomTable/farmdata2_modules/fd2_tabs/fd2_example/ui.spec.js) file to repair testing coverage integrity to the page. 
- [x] The [seedingReport.html](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/updatedCustomTable/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html) file needs to be refactored to make use of the new CustomTableComponent. This means modifying how data is sent to the component and potentially modifying/deleting functions related to props that were deleted in the CustomTableComponent. ~~Alongside this the [SeedingReport.spec.js](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/updatedCustomTable/farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.spec.js) file needs to be modified to repair full page testing coverage.~~
- [x] The [transplantingReport.html](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/updatedCustomTable/farmdata2_modules/fd2_tabs/fd2_barn_kit/transplantingReport.html) file needs to be refactored to make use of the new CustomTableComponent. This means modifying how data is sent to the component and potentially modifying/deleting functions related to props that were deleted in the CustomTableComponent. ~~Alongside this the [transplantingReport.spec.js](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/updatedCustomTable/farmdata2_modules/fd2_tabs/fd2_barn_kit/transplantingReport.spec.js) file needs to be modified to repair full page testing coverage.~~
- [x] Create issues for the end-to-end Cypress tests for the Seeding Report and Transplanting report. This isn't restrict solely to myself, if someone would like to contribute to that that would be much appreciated. If not, I will get around to it eventually.
---
__Licensing Certification__


farmdata2_modules/fd2_tabs/resources/NewCustomTableComponent.js
```diff
@@ -0,0 +1,352 @@
...
+                <tbody>
+                    <tr class="table-text" 
+                    v-for="(row, ri) in rows"
+                    :data-cy="'r'+ri">
+                        <td v-for="(item, ci) in row.data"
+                        v-if="columns[ci].visible"
+                        :data-cy="'td-r'+ri+'c'+ci"
+                        style="text-align:center">
+                            <div v-if="rowToEditIndex!=ri || columns[ci].inputType.type == 'no input'"
+                            :data-cy="'r'+ri+'c'+ci"
```
braughtg, line 80: This is the part that is rendering the table row if the row is not currently being edited, or if it does not have an input element type.  So when the row is not being edited, the check box element is not being rendered (same for the button.). This was fine when all elements were just text when they were not being edited.  But now that there are some that are not text (e.g. checkbox, button), you'll need to generalize this or add a special case to handle those elements for when a row is not being edited.

braughtg:
__Current Issues__
~~Checkboxes/buttons are only rendered in when you edit a table row. Otherwise the HTML text fed into the table's row data is displayed. As shown below.~~

1. `RegexInputComponent` is running into major errors. For starters it won't save a number, it'll retain its original value. There's a problem where the regex comparison isn't working properly (e.g. the regex for positive ints only allows '4.5') and when trying to add `@input-changed` and `@match-changed` for checking if the value is valid there's an error thrown into the console that says e.apply is not a function. 

|image |
|:--:|
| *Positive integer regex allows decimal numbers*|

|image |
|:--:|
| *Error thrown when I try to incorporate match-changed and input-changed into the html template*|

https://user-images.githubusercontent.com/31524934/192862575-1899acc6-57bc-43c3-a584-a2ad7c961cc8.mp4
|  |
|:--:|
| *Video showing that putting a value in doesn't actually change the value*|

braughtg:
The issue with the input value not changing the value displayed suggests that the row-edited event is not being emitted correctly or that page using the table is not handing the row-edited event properly.  I would check to ensure that the event is being emitted, and then if so, ensure that the page using the table is handling it correctly.  That event handler in the page will need to change the original row data from which the prop to the table is generated.  In the example page, I think that is just an array that is being used.  In something like the seeing report, that will be the original data from the api call.  The change to that data will be picked up by vue because the rows passed to the table is a computed property.  When that computed property changes, the prop will change, when the prop changes, the values displayed in the table should also change.

FutzMonitor:
By observing the original `CustomTableComponent` I noted that when you change the value inside of the an input box of type number the value being passed becomes a string (inside of the Vue devtools you can see quotation marks surrounding the newly inputted value). 

As the author of the `RegexInputComponent` I am very much aware that strings being inside of the component's input box when that input box's type is number can cause some serious problems: 

image
*Image of the value being emitted back*

The value emitted back is empty but the box goes back to being white indicating a valid input to the user. Ultimately the value doesn't get save, see video above for example, but I have suspicion that passing a string is causing some critical problems.

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
```diff
@@ -334,11 +378,13 @@ <h1 class="text-center">Seeding Report</h1>
...
                  */
                 deleteRowLog(logID) {
-                    let url = '/log/' + logID
-                    deleteRecord(url, this.sessionToken)
-                    for(var i=0; i < this.seedingLogs.length; i++){
-                        if(this.seedingLogs[i].id === logID){
-                            this.seedingLogs.splice(i, 1)
+                    for(let i = 0; i < logID.length; i++) {
+                        let url = '/log/' + logID[i]
+                        deleteRecord(url, this.sessionToken)
```
braughtg, line 383: deleteRecord returns a Promise.  Here the deleting of the local rows should only happen if the promise resolves successfully.  So that code should go into the then().  There should also be a catch for the error in case there is an issue with deleting the record from the DB.  Use the other API calls where we display the error banner as a guide.

farmdata2_modules/fd2_tabs/fd2_barn_kit/transplantingReport.html
```diff
@@ -234,11 +231,13 @@ <h1 class="text-center">Transplanting Report</h1>
...
                  */
                 deleteRowLog(logID) {
-                    let url = '/log/' + logID
-                    deleteRecord(url, this.sessionToken)
-                    for(var i=0; i < this.transplantingLogs.length; i++){
-                        if(this.transplantingLogs[i].id === logID){
-                            this.transplantingLogs.splice(i, 1)
+                    for(let i = 0; i < logID.length; i++) {
+                        let url = '/log/' + logID[i]
+                        deleteRecord(url, this.sessionToken)
```
braughtg, line 236: Same comment as above applies here.

farmdata2_modules/fd2_tabs/fd2_example/ui.html
```diff
@@ -304,34 +312,52 @@
         button2Clicked() {
             this.leftButtonDisabled = false
         },
-        deleteTableRow(rowID) {
-            for(var i=0; i < this.tableData.length; i++){
-                if(this.tableData[i].id == rowID){
-                    this.tableData.splice(i, 1)
+
+        deleteTableRow(rowIDs) {
```
braughtg, line 316: Here too.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
```diff
@@ -413,13 +413,14 @@ let CustomTableComponent = {
...
-        //     handler(newArr) {
-        //         this.updatedVis = newArr
-        //     },
-        //     deep: true
-        // }
+        columns: {
+            // using deep watch to track nested property changes
+            handler(newArr) {
+                console.log("Hey!")
+                this.columns = newArr
```
braughtg, line 420: I'm not sure that resetting the prop internally here does what we need to do.  The original code was updating an element of the data, not the prop.  So maybe there is something going on with that?   It would be worth doing a little reading on deep watches and how they are typically used.

braughtg: You may want to put the prop into a data element (as was done with `updatedVis` in the original code).  Then in the watch, update that data element.  Then use that data element in the `v-if`s in the html.

braughtg: Looks like you can also watch for changes just in specific nested properties inside of a prop.  See: https://stackoverflow.com/questions/42133894/vue-js-how-to-properly-watch-for-nested-data

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
braughtg: @FutzMonitor Have you look into handling possible errors during the delete here and at the other places I marked it?

farmdata2_modules/fd2_tabs/fd2_example/ui.html
```diff
@@ -396,11 +396,18 @@
...
+        cloneEvent(IDsToClone){
+            for(let i = 0; i < IDsToClone.length; i++){
+                for(let j = 0; j < this.newTableData.length; j++){
+                    if(this.newTableData[j].id == IDsToClone[i]){
+                        clonedRow = structuredClone(this.newTableData[j])
+                        // Prevents buttons sharing same row.id and thus clicking one
+                        // ticks another one 
+                        clonedRow.id = this.newTableData.length + 1
+                        this.newTableData.push(clonedRow)
+                        break
```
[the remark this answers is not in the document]
braughtg: The `break` is not exiting the conditional, but is exiting the inner `j` for loop.  So it is advancing to the next element of `IDsToClone` without searching the rest of the `tableData`.  This makes sense because it has already found the row.

I am unsure as to why without the break that the `j` look wouldn't just continue until the full `tableData` was searched and then go onto the next `IDsToClone` element and repeat.  

I think this is fine as it is.  You could convert to using a `filter` in place of the inner `j` loop if you prefer.

farmdata2_modules/fd2_tabs/fd2_barn_kit/seedingReport.html
braughtg: I think having the error banner appear is sufficient.  I think I may have just missed the fact that the delete was moved inside the loop and it does now have a then and a catch on it.   So this looks good to me now.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
braughtg: Agreed.  Leave it as it is for now and add an issue to make the dropdowns in the seeding report have context sensitive contents based on the type of seeing etc... You'll need to think about it a little in how to write up the issue to describe how the dropdown contents should be computed.  You can also then in a comment on that issue describe the approach to fixing this that we have discussed (i.e. making the prop a computed property rather than the data element).

FutzMonitor: I believe this would fall under #585

braughtg: Similarly, there should be a new issue created that has the edit-clicked event emit the id of the row that is being edited as its payload.

FutzMonitor: That is also mentioned under #585

braughtg: I see that you already have an issue filed that captures both of the above changes.  Thanks!

farmdata2_modules/fd2_tabs/resources/NewCustomTableComponent.js
braughtg: I'm not sure I fully understand the issue here.  But from what you say, it seems that the buttons and check boxes are not disabled when editing a row unless the row is blank?  Is that accurate?   That doesn't seem like the behavior we want.  We'll want to sort this out if it is still not behaving correctly.  Please update on the current status.

FutzMonitor: This shouldn't be the case anymore. This was when custom buttons were still part of the table row but has been moved to the top of the table. The radio buttons have been updated to be disabled at all times except for when that row is being edited. As far as I remember, rendering isn't an issue anymore.

braughtg (approved):
Looks good!  Great work getting this big task together!

PR #591, opened 2022-12-12
Updating Export Button Functionality

### LINKED ISSUE #571: Export Button: Update Functionality.
*This issue is to be visited once #530 has been resolved.*

# Current Design
Clicking on the export button for a table exports all visible data in the table. 
However, refactoring of the `CustomTableComponent` has included a column of checkboxes that can be used to select which rows we want printed into the csv and which we do not. To keep the functionality of the CSV export button consistent with those of the custom buttons, it should also have its functionality tied to those checkboxes.

# Desired Design
1. Make the necessary changes to the `CustomTableComponent`'s csv exporting function [[here](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js#L231)] so that it respects the checkboxes and only exports those rows whose boxes are checked.
2. Update the [CustomTableComponent.spec.comp.js](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/resources/CustomTableComponent.spec.comp.js) to test this new capability of the csv export button within the table.

### LINKED ISSUE #589: Table Export includes `-` characters at end of Comments field.
When a report is exported using the export button, values in the comments column include extra `-` characters at the end of the comment.  

**To Replicate**
1. Go to BarnKit / Seeding Report
2. Search for 05/01/2020 - 05/05/2020
3. Click the Export Button above the table.
4. Open the resulting CSV file.
5. Observe the contents of the Comments Column
6. Note extra `-` that do not appear in the Comments column in the table itself in FarmData2.

FutzMonitor:
__Pull Request Description__

Closes #589 and Closes #571 . This PR aims to update some functionality of the export button that were left out when the CustomTableComponent was being reworked. This includes updating its behavior to be consistent with the behavior of other buttons that are part of the table. Additionally, this PR addresses a bug where the export button leaves behind some `-` in the comments field. 

**This PR is ~~not~~ ready to be merged into upstream/main**
Below are some things that need to be completed before this PR is ready for review: 
- [x] Have a larger conversation about how the checked rows should behave after a button action (consistency issue) 
- [x] Unit test needs to be guarantee test coverage of updated export functionality 

---
__Licensing Certification__


FutzMonitor:
@braughtg Do you believe all the buttons should reset the `indexesToAction` array? That is to say that when I delete, export, or commit some custom action that afterwards the rows that were checked are now unchecked like it's preparing itself for the next set of instructions? I admit that the delete button is a bit of a special case, but I think maybe we should reset them for all buttons and have them all have that behavior.

braughtg:
> Do you believe all the buttons should reset the `indexesToAction` array?

This sounds like a good idea.  It is a clear rule on how they should behave and should be intuitable by the user.  If it turns out to be inconvenient and we hear feedback otherwise from end users, we can adapt.  

So move forward under this assumption.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
```diff
@@ -364,10 +364,10 @@ let CustomTableComponent = {
 
             for(let i = 0; i < this.rows.length; i++){
                 for(let j = 0; j < this.columns.length; j++){
-                    if(this.columns[j].visible){
+                    if(this.columns[j].visible && this.indexesToAction.includes(this.rows[i].id)){
                         if(typeof this.rows[i].data[j] === 'string'){
                             cleanHTML = this.rows[i].data[j].replaceAll(/(<p[^>]+?>|<p>|<\/p>|<br \/>)/img, "")
-                            cleanHTML = cleanHTML.replaceAll(/(\r\n|\n|\r)/gm, "-")
```
braughtg, line 370: I seem to recall that there was a reason we were using `-` and `--` here.  I think it had to do with inserting new lines or columns not aligning properly in the CSV.  Double check the output that is generated to ensure that just using here doesn't mash stuff together and make it unreadable.  Also be sure to understand when `-` and `--` are used so that we address all instances of issues, even if they don't show up in the simple example produced by the steps in the issue.

FutzMonitor:
Do you have any idea what may be causing this [unrelated] in the Seeding Report? @braughtg 
image

It doesn't do this on my laptop. I tried rebuilding the database and that also doesn't work. It seems to be a bug only I'm running into even though my main is exactly like the upstream/main.

braughtg:
Maybe a browser specific issue?

On Dec 13, 2022, at 10:40, Chris ***@***.***> wrote:

﻿
Caution:This email originated from outside of the College. Do not click links or open attachments unless you recognize the sender and know the content is safe. Report anything suspicious to the HelpDesk.


Do you have any idea what may be causing this [unrelated] in the Seeding Report? @braughtg<https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fbraughtg&data=05%7C01%7Cbraught%40dickinson.edu%7Ce33b45bc3d9d4ca96d4008dadd204a91%7C6232b05576b94c139b88b562ae7db6fb%7C0%7C0%7C638065428019839563%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=bsSQz1CGxTVVqUgsQJerMzA8IrOMFAa7Bm4w%2FUOs%2Fss%3D&reserved=0>
[image]<https://nam12.safelinks.protection.outlook.com/?url=https%3A%2F%2Fuser-images.githubusercontent.com%2F31524934%2F207377306-a77532b3-1f6e-4d45-90db-176024fe7f34.png&data=05%7C01%7Cbraught%40dickinson.edu%7Ce33b45bc3d9d4ca96d4008dadd204a91%7C6232b05576b94c139b88b562ae7db6fb%7C0%7C0%7C638065428019839563%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=RGHWe%2FFquGSBxH%2Bblxx6JVmToDSzQfk0WppKCo5w9Zg%3D&reserved=0>

It doesn't do this on my laptop. I tried rebuilding the database and that also doesn't work. It seems to be a bug only I'm running into even though my main is exactly like the upstream/main.

FutzMonitor:
> Maybe a browser specific issue? 

Seems to be, it works just fine on Chromium based browsers (tested on Opera GX, Chrome, and Edge).

FutzMonitor:
Could you pull this PR recreate these steps on your machine? @braughtg 

0. Open up any Chromium based browser (Chrome, Opera)
1. Visit the Seeding Report under the Barnkit tab.
2. Simply generate the report with the preset dates
3. Select All the rows and export a CSV. 
4. Open the CSV and observe that it stops at _cabbage_ on the 26th of June instead of the final row which is a beet. 
5. Repeat steps 1-4 but now within FireFox and observe that Firefox doesn't exhibit the same behavior. 

I'd like to know if my machine is being buggy or if the export functionality isn't being consistent across browsers. If the latter is the case, this is really concerning because Chromium browsers do not play nice with Vue dev tools which makes it challenging to debug Vue code.

braughtg:
> > Maybe a browser specific issue?
> 
> Seems to be, it works just fine on Chromium based browsers (tested on Opera GX, Chrome, and Edge).

This worked fine for me in Chrome and in Firefox on MacOS, and in Firefox in the Dev Container.

braughtg:
> I'd like to know if my machine is being buggy or if the export functionality isn't being consistent across browsers.

I saw the same behavior in both Chrome and Firefox on MacOS.  The CSV terminated with the 2020-06-26 row for Cabbage when used from both browsers.  So neither browser produced the full report.

In Firefox, in the Dev Container I'm getting the abbreviated columns (i.e. mobile view?) where there is no row select check box and thus no way to select all of the rows or to delete or edit rows.  This happened no matter what the size of the window was, so this may be a separate issue.  But, when I clicked the export button, it exported the whole table.  But with fewer columns per row, becuase that is what was being displayed.

So, the issue with the export being cut off seems like it might be due to be a limit on the size of the data that the CSV library you are using.  Is that possible?

FutzMonitor:
> So, the issue with the export being cut off seems like it might be due to be a limit on the size of the data that the CSV library you are using. Is that possible?

It seems like this might be it. I uninstalled and reinstalled FireFox on my machine and that seemed to fix the problem I was having before. Now FireFox is behaving in-line with the Chromium based browsers where the export only exports up to a certain number of rows.

In regards to the mobile view appearing in the dev container, I did not experience similar issues. I got full columns with nothing excluded and had the checkboxes present as well. That should be something we keep an eye out for in case it's no machine dependent and is an independent bug. 

image
*dev container is presenting normally on my end*.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
FutzMonitor: This seems to be guaranteed. I didn't see any problems when I made my changes on my end.

braughtg: I think we need to revert this change and continue replacing all new line characters with a `-`.  Otherwise what is happening is that any comments that contain new lines to separate items or phrases are being mashed together with no space between the items.  You can observe this in the second comment in the seeding report generated using the default dates.  In the rendered table in FarmData2 the comment is broken into multiple lines.  In the exported CSV the lines are mashed together with no spaces.  We had added the `-` just to keep the items visually separated, even though they will be on a single line in the CSV.  We might consider using " - " a dash with a space on either side instead to make it even a little better.

braughtg (changes requested):

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
FutzMonitor: This has been made the case in the latest commits.

braughtg (approved):

PR #593, opened 2022-12-17
CustomTable: Edit-Clicked Event Has Row.id Payload

### LINKED ISSUE #588: CustomTableComponent: Edit-Click Emits Row Data
## Current Functionality
Currently when the CustomTableComponent's edit button is clicked it emits an event called _edit-click_ which can be used to run whatever methods needed on the parent page. For example, the Seeding Report page uses this emitted event to disable the filters on the page so that users cannot alter the filters while editing a row. However, the emitted event does not emit any payload. 

```        editRow: function(index){
            this.rowToEditIndex = index
            this.currentlyEditing = true
            this.originalRow = JSON.parse(JSON.stringify({ 
                'id': this.rows[index].id,
                'data': this.rows[index].data
            }))
            this.editedRowData = { 
                'id': this.rows[index].id,
                'data': this.rows[index].data
            }
            this.$emit('edit-clicked')
        },
```
_As can be seen, an event is emitted but no payload along with the event._
## Desired Functionality
The _edit-click_ emitted event needs to emit the row data of the row being edited. This will require editing the _editRow_ method [[here](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js#L173)]. Additionally, you will need to add a test case in the CustomTable's Cypress test [[here](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/resources/CustomTableComponent.spec.comp.js)] that assures that this payload is sent is correct.

FutzMonitor:
__Pull Request Description__

Closes #588. This PR adds `row.data` to the payload of the the `editRow` function inside the CustomTableComponent. This will allow parent pages to have knowledge on the table's current editing and allow it to update anything on its end (like filters) inside to make sure only the appropriate options for editing are available. 

**This PR should be merged after the export functionality PR has been merged**
I made the mistake of branching off from that branch instead of main so that means that this PR has those changes as well. I'll leave this as a draft for the time being until that one is merged. 

---
__Licensing Certification__


farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
```diff
@@ -364,10 +364,10 @@ let CustomTableComponent = {
 
             for(let i = 0; i < this.rows.length; i++){
                 for(let j = 0; j < this.columns.length; j++){
-                    if(this.columns[j].visible){
+                    if(this.columns[j].visible && this.indexesToAction.includes(this.rows[i].id)){
                         if(typeof this.rows[i].data[j] === 'string'){
                             cleanHTML = this.rows[i].data[j].replaceAll(/(<p[^>]+?>|<p>|<\/p>|<br \/>)/img, "")
-                            cleanHTML = cleanHTML.replaceAll(/(\r\n|\n|\r)/gm, "-")
+                            cleanHTML = cleanHTML.replaceAll(/(\r\n|\n|\r)/gm, "")
```
braughtg, line 370: Here we should leave this as a `-` or maybe a " - " so that multiple lines do not get mashed together.  See the comment in the first row of the Seeding report table generated using the default dates for an example.  There each item should be separated with `-` or `" - "`

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
```diff
@@ -296,7 +296,7 @@ let CustomTableComponent = {
                 'id': this.rows[index].id,
                 'data': this.rows[index].data    
             }
-            this.$emit('edit-clicked')
+            this.$emit('edit-clicked', this.rows[index].data)
```
braughtg, line 299: This emits the full row data.  Does it make sense to emit just the `id` or is there value in emitting all of the data in the row?

braughtg (changes requested):

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
FutzMonitor: This can be disregarded as it's primarily being addressed in PR #591. I just unfortunately made this feature branch from that one and not from main meaning those file changes can be found here as well.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
FutzMonitor: Either way, I believe it's a point of reference. If we emit the `id` here, then the parent page will have to take the time to find the log within its own rows array and grab that data there to determine what should be filtered. Would you like that to be something the parent page handles or the child component?

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
braughtg: Sounds good. Though this now has created a conflict that will need to be resolved before this can be merged.  Just resolve that conflict accepting your prior changes and push the update.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
braughtg: I agree that since we have the data readily available it seems sensible to emit it so that the page doesn't have to look it up.  But on the other hand, there may be cases where the page will need to lookup the full log for something that doesn't appear in the `data`.   Is there something in `this.rows[index].data` that will allow the code in the page to look up the corresponding full log?  Or is that what the `id` does?

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
```diff
@@ -89,7 +89,7 @@ let CustomTableComponent = {
                 <thead>
                     <tr class="sticky-header table-text" data-cy="table-headers">
                         <th
-                        v-if="canDelete || customButtons.length > 0"  
+                        v-if="canDelete || customButtons.length > 0 || csvName != ''"
```
braughtg, line 92: Would it make sense to use a computed property for this so that we can ensure that it is consistent across all elements?

FutzMonitor: I believe we have all the cases covered now and a computed property would really just encapsulate these cases to make it readable for future developer. It might be worthwhile implementing it then if it makes it easier to abstract these cases and make them more manageable and extendable in the long run instead of having to look for each of these locations whenever a new case is created.

However, that implementation goes beyond the scope of this issue. A new issue would have to be created for that proposed change.

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
FutzMonitor: In that case, it would make more sense to have it simply emit the `id` and have the parent page handle lookup of the log and grab whatever information it needs. Unfortunately the `this.rows[index].data` only emits that logs data without any way of quickly fetching that log from the rows array in the parent page. Not to mention even if we did workaround that issue two logs might have identical data but have different id's. So it is definitely better to use the `id` in this case and just allow the parent page to handle it. 

I didn't grab from `editedRowData` because I think that might cause problems down the line. Currently we want to populate a dropdown based on the seeding type of the log. Fortunately, seeding type is not an editable data property of a log, but it may be the case down the road where we want to populate some property and the data property we want to look at can be edited which might update the data property options. Is that something we want or do we want it to be set in stone once you hit the edit button?

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
braughtg: Sounds reasonable.  Would you create that issue?

farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js
braughtg: Agreed.  Let's just emit the `id` then.  The page can then look up the full log and obtain whatever data that it needs from there.

FutzMonitor (commented):
Everything should be good now.

braughtg (approved):

farmdata2_modules/fd2_tabs/fd2_example/ui.html
```diff
@@ -291,7 +291,7 @@
             this.cropList.push('ZUCCHINI')
         },
         dateChange(newDate) {
-            this.chosenDate = newDate
+            this.chosenDatableDatate = newDate
```
braughtg, line 294: Looks like a typo snuck in here.

braughtg (changes requested):

farmdata2_modules/fd2_tabs/fd2_example/ui.html
FutzMonitor: This should be fixed now.

braughtg (approved):
This looks good.  

I am in the midst of a bit reorganization of the repo.  The files where these changes are will have moved.  So, when I have completed the reorg, I will integrate these changes in to the corresponding files in their new locations and credit you as co-author on the commit.

braughtg:
This is a note to me from me...  Add documentation for the payload of the edit-clicked event to the JSDoc for the component.

braughtg:
Changes here were included in #604.

PR #598, opened 2022-12-19
Set permissions on .fd2gids for WSL/Linux

### LINKED ISSUE #584: Dev Container Fails to Launch Unless Removed from Local Storage
## Description of the Issue
The dev container fails to launch with an error message when it is launched without having it removed from the local storage after the previous launch. 
image

It runs successfully if the container is removed from the local storage after the previous run (by manually deleting the container from the docker desktop or running the `docker rm` command).

## Steps to Reproduce

1. Run Farmdata with `./fd2-up.bash` inside the docker directory. 
2. Farmdata should launch without error as the dev container did not exist in the local storage when Farmdata was launched for the first time. 
3. Stop Farmdata with `./fd2-down.bash`. All containers should be removed except the dev container as the fd2-down.bash script does not remove the dev container.  
image
4. Relaunch Farmdata `./fd2-up.bash`. The dev container fails to launch with error messages. 
image
## Suggested Solution

Temporary solution: remove the dev container every time by including `docker rm fd2_dev` in the fd2-down.bash script file.

braughtg:
__Pull Request Description__

Changes the permissions on the `.fd2gids` directory and files that are created by the `fd2_up.bash` script so that they are writeable by the `fd2grp` group.  It is hoped that this will allow the `.fd2gids` directory to be successfully mounted into the dev container on subsequent starts.

Closes #584

---
__Licensing Certification__

### DIFF
--- a/docker/dev/startup.bash
+++ b/docker/dev/startup.bash
@@ -21,11 +21,13 @@ echo "fd2dev" | sudo -S usermod -a -G $HOST_DOCKER_GID fd2dev
 echo "fd2dev" | sudo -S chgrp +$HOST_DOCKER_GID /var/run/docker.sock
 echo "fd2dev" | sudo -S chmod 775 /var/run/docker.sock
 
-# Ensure that the fd2grp has RW access to mounted FarmData2 directory.
+# Ensure that the fd2grp has RW access to mounted FarmData2 and .fd2 directories.
 # This probably is not necessary as in linux the group will reflect
 # the host group.  But it does make things consistent across macOS and linux.
 echo "fd2dev" | sudo -S chgrp -R fd2grp /home/fd2dev/FarmData2
 echo "fd2dev" | sudo -S chmod -R g+rw /home/fd2dev/FarmData2
+echo "fd2dev" | sudo -S chgrp -R fd2grp /home/fd2dev/.fd2
+echo "fd2dev" | sudo -S chmod -R g+rw /home/fd2dev/.fd2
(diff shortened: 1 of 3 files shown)


FutzMonitor:
> Please pull this PR and try bringing fd2 up and down a few times to see if the changes resolve the issue you have been having with the dev container (#584). Report back here what you observe. Thanks!

@braughtg Unfortunately this didn't work on my end. I removed my dev container first to allow it to build from the image. After bringing it down and bringing it back up once, it outputted basically the same mounting error from before:
image

Here are the permissions again: 
image

braughtg:
@FutzMonitor Thanks for giving that a try.  Based on what you reported, I made another change.  Could you give it another try when you get a chance?  Thanks!

FutzMonitor:
@braughtg Just pulled the branch changes and ran it again. Made the same actions: delete the dev container before running the up script, down script, and it once again outputted the failed to mount error: 
image

Here are the permissions again: 
image

FutzMonitor:
@braughtg I removed the containers, volumes, and images and ran the up script. Then I ran the down script and up and it worked. When I tried that again though, down and up again, the mounting error was outputted: 
image

Here are the permissions: 
image

braughtg:
@FutzMonitor Thanks.  This one has me stumped!!  Looking like I'll go with the fallback of just removing the dev container like all of the others.

braughtg:
It seems that this approach is not going to work.

Issue #585, opened 2022-12-08
Seeding Report: Filtering Area/Crop Name Arrays Depending on Seeind Log


FutzMonitor:
## Current Functionality
When you edit a row in the Seeding Report page all crops and areas are listed in the dropdowns. This doesn't make sense because it allows users to place a tray seeding in a field area while also allowing a direct seeding change the area to designated tray seeding spaces. Additionally, some crops are exclusively tray/direct seeded, so users shouldn't be allowed to change a tray seeding to a crop that's exclusively direct seeded and vice-versa for direct seeding. 

## Desire Functionality
Some filterable functionality needs to be added into the computed methods of the Seeding Report page that changes the dropdown options according to the log's seeding type. If it's a direct seeding only display the possible, relevant options for direct seedings and vice-versa for tray seeding. This will require changes to the [CustomTableComponent.js](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/resources/CustomTableComponent.js) which will necessitate updates to the [CustomTableComponent.spec.comp.js](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/resources/CustomTableComponent.spec.comp.js) to test this new functionality. The changes would be adding a payload to the `edit-clicked` emit so that it emits the `id` of the log being edited. Then the parent page can determine the options of the dropdown based on some information located in that log's data (in this case, the seeding type).

## Greater Issue
This issue deals with some of the questions asked in issue #466. 

**What should be editable?** 
This applies to many things in the project: columns in the report pages, the flow of editing (i.e when a planting log is changed does its associated logs update? Should you even be able to update planting logs?). There are many questions surrounding editing that need to be addressed as a whole moving forward so that we follow good practices in pages and avoid larger problems of creating inconsistent logs.

braughtg:
Potentially split this issue into two.  One for adding the payload to the `edit-clicked` event and testing that.  This can be done independent of any other changes as the payload will just be ignored by current code.  Then a second issue, for using that payload in the seeding report page to make the dropdown menus for area and crops context sensitive.

FutzMonitor:
The `edited-click` changes mentioned in the issue are no longer necessary to fulfill this issue but are necessary to complete it. That is, this issue should follow after the closing of #588.

FutzMonitor:
[...]

Issue #466, opened 2022-06-07
Seeding Report Editing Issues/Bugs/Concerns


braughtg:
Which columns in the seeding report should be editable and when?

* It seems to make sense that some columns should be editable and some may not be. 
  * For example, you can edit hours but not the type of seeding.
  * When a column that needs to be changed (e.g. type of seeding) is not editable the solution is to delete the log and recreate it.
* Seems like some things should be able to be edited or deleted only under certain conditions. 
  * What happens if a transplanting or harvesting log already exists that is associated with the seeding?
  * This needs to be researched.  Perhaps if the planting is changed then everything will be alright? 
  * Note however that the crop_tid field added to the seeding logs to make it unnecessary to fetch the plantings will also need to be updated if the crop is edited.
* When editing the location of a seeding the options should be consistent with the seeding type.
  * Only greenhouses for tray seedings and only field or bed locations for direct seedings (as in the input form)
*  When editing fields that are constrained (e.g. hours or workers) the edit field should be a RegexInputComponent with an appropriate regular expression to prevent invalid inputs.

FutzMonitor:
It seems that a separate issue should be opened to address what was discussed yesterday concerning an inflexibility of the CustomTableComponent: buttons. Every time a new button needs to be added to a table, the component shouldn't have to grow every time.

**A potential discussed solution to the button problem:** 
We remove all the buttons currently in the CustomTableComponent (save/cancel/edit/delete) and just add a general button prop that can be added to the end of the component. This increases the flexibility of the component because it reduces the burden of adding buttons to the component and puts that responsibility back on the page which should handle those button presses.  We would have to think about how one prop will account for many buttons, but I'm sure that wouldn't be too difficult.

braughtg:
Issue #530 addresses potential updates to the CustomTableComponent that may affect a solution to this issue.

FutzMonitor:
I believe it would be worthwhile to consult with Matt which elements of the logs should be editable from the report table. Here are a couple of bugs concerning edits and the report table that I have witnessed. 
1. Editing the row feet causes the bed feet to be recalculated. Normally this wouldn't be an issue, but it's misleading when the user changes both row feet and bed feet only for bed feet to ignore the user's input and recalculate itself. Perhaps it might be worthwhile disabling an element when one that affects its recalculation is being changed. 
2. Editing the date, area, and crop all currently do not update the log in the backend. 
[...]

PR #615, opened 2023-01-16
Backend and Frontend Fixes for Transplanting Report

### LINKED ISSUE #562: Transplanting Report: Editing Transplanting Logs Does not Update them in the Backend
## Current Behavior ##
1. Visit the Transplanting Sub-tab under the Barnkit module. 
2. Generate a report from 05/04/2020 to 05/04/2020.
3. Click the edit button and change the date, crop, area, trays, and user then click the save button.
4. Visit Hoppscotch and using the GET function enter this link in the URL: `http://localhost/log/1183`
5. Verify that the data that you changed inside of the Transplanting Report has not changed (date changes timestamp, area changes can be seen inside of movement, crop changes can be see at the bottom with the data crop_tid, and user can be found inside of the log owner array). 

image
*Changes on the frontend are visible*

image
*Changes have not actually been made in the backend (trays for example is still 8 despite being 60 in the report)*

## Desired Behavior ##
The desired behavior would be that the user can change the date, crop, area, trays, and user and those changes are actually reflected in the backend. Currently the user is allowed to change the value inside of _Bed Feet_, but it might not be worth allowing them to do this because the backend doesn't have a quantity for this anyway. Additionally, changing Row Feet recalculates Bed Feet anyway. It would be worth testing if the transplanting report, after fixing the backend updates, suffers from similar bugs that the Seeding Report suffers from. This includes:
1. Changing an area and a quantity causes a 422 status code
2.  Making sure that changing the quantity along with date, crop, area, comments, and user actually updates the frontend appropriately (Seeding Report had an issue where frontend was not being updated correctly when a quantity and another element were changed at the same time)

The Cypress e2e file for this should also be updated with a more rigorous testing suite. 

## Afterword
Pull request #551 is a good place to look into for people seeking to solve this issue. It confronts the exact same issue except for the Seeding Report.

FutzMonitor:
__Pull Request Description__

Closes #562. This issue makes some changes to the `updateRow` function located in the Transplanting Report HTML file [[here](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/fixTransplantingBackend/farmdata2/farmdata2_modules/fd2_barn_kit/transplantingReport/transplantingReport.html)]. These changes are not accompanied by any Cypress e2e testing as that is a separate issue to be addressed in issue #578. As a result, these changes have been tested by sight using Hoppscotch. 

Additionally, this PR makes a quick fix of the filters for both the Seeding Report and Transplanting Report where editing once would permanently disable the filters until the page was refreshed because the name of the edit cancellation event was changed and the pages were not updated accordingly to reflect this change. 

This PR also brings up the topic of #585 (and #466) of what should be edit-able. I disabled the ability to edit `bed feet` in the Transplanting Report because transplanting logs don't have quantity data for `bed feet` but also enabled the ability to edit `row/bed` because that does exist in the logs' quantity data. However, should users be allowed to edit that data? This is beyond the scope of this PR but wanted to quickly mention this quick thought that occurred to me whilst making this fix. 

---
__Licensing Certification__


braughtg:
I don't think we want the zipped database (`docker/db.sample.tar.bz2`) committed here.  Is there a reason that it needs to be?

farmdata2/farmdata2_modules/resources/CustomTableComponent.js
```diff
@@ -326,7 +326,6 @@ let CustomTableComponent = {
         },
 
         cancelRowEdit: function(){
-
```
braughtg, line 329: Maybe remove this change from this PR as well since it doesn't add anything?

FutzMonitor:
Those changes should be gone from this PR now.

PR #622, opened 2023-02-21
Generalization of the ErrorBannerComponent 

### LINKED ISSUE #601: Generalize ErrorBanner Component
There is an ErrorBannerComponent defined in the `resources` folder and demonstrated on the UI Subtab.  There should also be a SuccessBannerComponent that can be used to display success messages (e.g. "Log has been saved." ect...).  

One way to do this would be to generalize the ErrorBannerComponent so that a prop dictates if it is an ErrorBanner or a SuccessBanner.  Another prop can determine how long the banner is displayed or if it must be clicked to be dismissed.  This would help significantly with the consistency of the code and UI/UX across modules.

When this more general component is complete it should be fully integrated into all of the pages where it error and/or success banners are displayed.  Currently these are:
* Seeding Report
* Transplanting Report
* Seeding Input
* Config

### LINKED ISSUE #635: Typo in clearDupalCache bash Alias in dev container
In the `bash_aliases` file in `docker/dev` there is a typo in the definition of the `clearDrupalCache` command.

The command is currently declared as `clearDupalCache` (note the missing `r`).

The command should be declared as `clearDrupalCache`

To fix this:
* Edit the `bash_aliases` file.
* Rebuild the dev docker container to test it.
  * Note the name of the current dev image in Docker Desktop
  * Use `docker build` locally to create an image with exactly same name as the existing one.
* Use `./fd2-down.bash`
* Use `./fd2-up.bash` 
* Check that the `clearDrupalCache` command works in the dev environment.
* Make PR.
* The maintainers will rebuild the actual image using your fix and push it to Dockerhub.

FutzMonitor:
__Pull Request Description__
This PR closes #601. This pull request generalizes the functionality of the `ErrorBannerComponent` into the newer `BannerComponent` which can display error, success, and generic message banners. Additionally, the refactored component can support timeout (instead of clicking an _X_ to close the component, it'll collapse itself after 5secs) and is has a sticky style which makes it appear on screen without the intrusive auto scroll. 

This PR will not remove the pre-existing `ErrorBannerComponent` from the project. That is something that will need to be done in a separate PR that also replaces those instances with the new `BannerComponent`.

This PR is **NOT** ready to be merged into main. 
Prior to merging, the following items must be completed. 
- [x] Complete code review 
- [x] Remove all instances of the `ErrorBannerComponent` from the ui.html page. 
- [x] Update the end-2-end test for the ui.html page. 
- [x] Remove all lingering debugging code from the PR. This includes unnecessary comments.
- [x] ~~Regenerate documentation~~ (This is generated locally now)

Edit: Closes #635. 

---
__Licensing Certification__


farmdata2/farmdata2_modules/resources/BannerMessageConstants.js
```diff
@@ -16,6 +16,7 @@
  * </table>
  */
 var bannerMessageMap = {
+    standardError: {"msg": "Intermittent network error. Please try again later.", "class": "alert alert-danger alert-dismissible"},
```
FutzMonitor, line 19: Forgot to add this to the documentation at the top of the page. I'll be sure to do that in my next commit.

farmdata2/farmdata2_modules/resources/BannerComponent.js
```diff
@@ -0,0 +1,113 @@
...
+ * <tr><td>banner-handler</td>   <td>div containing bootstrap alert banner</td></tr>
+ * <tr><td>banner-close</td>   <td>'X' button that closes the alert banner</td></tr>
+ * <tr><td>banner-message</td>   <td>The message displayed inside the alert banner</td></tr>
+ * </tbody>
+ * </table>
+ * 
+ * <div>
+ *  @vue-prop {Object} [message={"msg": "Hello, I am a banner alert.", "class": "alert alert-warning"}] - Contains the message and class for the alert banner
+ *  @vue-prop {Boolean} [visible=null] - Set to true to show the alert banner. False makes the banner invisible.
+ *  @vue-prop {Boolean} [timeout=null] - Sets whether or not the banner dismisses itself which disables the 'x' to close. Set to false by default.
```
braughtg, line 17: If this is false by default should it say `[timeout=false]` instead of `[timeout=null]`?  That would also agree with the code below.

farmdata2/farmdata2_modules/resources/BannerComponent.js
```diff
@@ -0,0 +1,113 @@
...
+ * </div>
+ * 
+ * <div>
+ *  @vue-event visibility-update - Updates the visibility in the parent page when the child component has internal visibility updates. This means each page needs to implement a function that catches this emit and matches their visibility with the child's internal visibility. 
+ * </div>
```
braughtg, line 22: This component should have a code example that illustrates how a page would implement the handler.  See some of the functions in the `FarmOSAPI` documentation for examples of how to add code examples.  You don't necessarily need a complete example... but something a reader could easily adapt.  I.e. show the component with its prop bound and the event handler assigned, the data property and the event handler.

Also, when you say "Updates the visibility in the parent page..." that seems to suggest that this event does that.  However, the event doesn't do that.  It rather indicates that the banner has been hidden (I think... does this fire when the banner is shown too?).  So maybe a renaming here would make sense?   Call the event `banner-hidden` and then say something like "event indicating that the banner has been hidden either by the timeout or by the user clicking the close X.  The page should use this event to..."

farmdata2/farmdata2_modules/resources/BannerComponent.js
```diff
@@ -0,0 +1,113 @@
...
+    template: 
+        `<div data-cy='banner-handler' :class="bannerClass" 
+            ref="banner" role="alert" 
+            v-show="isVisible"
+            :style="{ position: ['-webkit-sticky', 'sticky'], top: [topNav]}">
+            <button data-cy='banner-close' type="button" class="close" 
+            aria-label="Close" 
+            @click="hideBanner" 
+            v-show="timeout==false">
+            <span aria-hidden="true">&times;</span>
```
braughtg, line 35: I'm a little unclear on the purpose of this line.

farmdata2/farmdata2_modules/resources/BannerComponent.js
braughtg, line 17: Should indicate the amount of time before the banner is dismissed here.

farmdata2/farmdata2_modules/fd2_example/ui/ui.html
```diff
@@ -1,6 +1,13 @@
 <!-- define a div to wrap around the app. -->
 <!-- apply v-cloak so that app is not visible until after being rendered -->
 <div id="ui" v-cloak> 
+    <!-- <div style="position: -webkit-sticky; position: sticky; top: 60px; background-color: forestgreen; padding: 3px;">Hello I am a DIV</div> -->
```
braughtg, line 4: Is this comment intended to still be here?

farmdata2/farmdata2_modules/resources/BannerMessageConstants.js
```diff
@@ -0,0 +1,35 @@
+// Declaring regular expression variables so that is
+// defined in all the pages served from out modules.
+
+
+/**
+ * Map of common banner messages used in the BannerComponent for displaying messages
+ * 
+ * <p><b>Banner Message Variables</b></p>
+ * <table>
```
braughtg, line 9: I think we can use in-code JSDOC comments to document these values instead of building a separate table.  See some of the code in the `FarmOSAPI.js` file and see if we can place comments with each message so that we don't  have to duplicate all of the content in the code and in the table in the comments.

farmdata2/farmdata2_modules/fd2_example/ui/ui.html
FutzMonitor: Strange, I thought I removed this with my latest commit.  That comment is not intended to be there. I'll remove it in the next commit.

farmdata2/farmdata2_modules/resources/BannerComponent.js
```diff
@@ -0,0 +1,113 @@
...
+ * <thead><tr><th>Value</th>        <th>Description</th></tr></thead>
+ * <tbody>
+ * <tr><td>banner-handler</td>   <td>div containing bootstrap alert banner</td></tr>
+ * <tr><td>banner-close</td>   <td>'X' button that closes the alert banner</td></tr>
+ * <tr><td>banner-message</td>   <td>The message displayed inside the alert banner</td></tr>
+ * </tbody>
+ * </table>
+ * 
+ * <div>
+ *  @vue-prop {Object} [message={"msg": "Hello, I am a banner alert.", "class": "alert alert-warning"}] - Contains the message and class for the alert banner
```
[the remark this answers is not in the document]
FutzMonitor: I can add that wording into the documentation to have a clear indication of where future devs can look for if they want to alter this component.

farmdata2/farmdata2_modules/resources/BannerComponent.js
FutzMonitor: This is my mistake. Originally the default is set `null` but that has changed since then. I'll make sure to update it in the next commit.

FutzMonitor: I'll be sure to add that to the documentation.

farmdata2/farmdata2_modules/resources/BannerComponent.js
FutzMonitor: You're correct, This emitted event only emits when the banner is hidden, so it would be better to call the event `banner-hidden` since it only fires when the banner is hidden. I don't believe there needs to be any update when the visibility is updated to true on the parent side since that'll go down into the child component. Then the hiding of the banner is handled by the child which communicates with the parent to inform it that the banner component is now hidden and its visibility value should now be false. 

I'll look into the `FarmOSAPI` and make sure to provide an example.

farmdata2/farmdata2_modules/resources/BannerMessageConstants.js
FutzMonitor: Sounds good, I'll look into it.

farmdata2/farmdata2_modules/resources/BannerComponent.js
FutzMonitor: I believe there was some CSS reason for the inclusion of this element. I'll look into it and see if it can be removed.

FutzMonitor (commented):
I need to make these changes first, but I'll flip my PR back to a draft in the meantime.

farmdata2/farmdata2_modules/resources/BannerComponent.js
FutzMonitor: This has to be left in there because it actually displays the 'x' that needs to be clicked on.

FutzMonitor:
I believe a good amount of issues have been addressed. My primary concern now is how the documentation looks, so I'll await your feedback on that one. @braughtg

farmdata2/farmdata2_modules/resources/BannerComponent.js
```diff
@@ -0,0 +1,120 @@
...
+ *  @vue-prop {Boolean} [visible=null] - Set to true to show the alert banner. False makes the banner invisible.
+ *  @vue-prop {Boolean} [timeout=false] - Sets whether or not the banner dismisses itself which disables the 'x' to close. Set to false by default. Currently set to self-dismiss after 5 seconds.
+ * </div>
+ * 
+ * <div>
+ *  @vue-event visibility-update - Updates the visibility in the parent page when the child component has internal visibility updates. This means each page needs to implement a function that catches this emit and matches their visibility with the child's internal visibility.
```
braughtg, line 21: Looks like this event has changed names.  The Docs should be updated to reflect that.

farmdata2/farmdata2_modules/resources/BannerComponent.js
```diff
@@ -0,0 +1,120 @@
...
+ *  @vue-prop {Boolean} [timeout=false] - Sets whether or not the banner dismisses itself which disables the 'x' to close. Set to false by default. Currently set to self-dismiss after 5 seconds.
+ * </div>
+ * 
+ * <div>
+ *  @vue-event visibility-update - Updates the visibility in the parent page when the child component has internal visibility updates. This means each page needs to implement a function that catches this emit and matches their visibility with the child's internal visibility. 
+ *  @returns false because the banner has been dismissed (either via clicking the x or the timeout).
+ *  @example <caption>Retrieve all planting assets into a new array.</caption>
+ * updateBannerVisibility(newBool) {
+ * this.bannerVisibility = newBool 
+ * // For a specific example make sure to check the UI page's source code.
```
braughtg, line 26: I would like to see a code example here that shows how to respond to the `banner-hidden` event.  That piece is something that will be easy to mess up and should be documented.

braughtg (changes requested):

FutzMonitor (commented):
Documentation has been updated.

FutzMonitor:
This is what documentation looks like for the BannerComponent. 
image

I'll be sure to add the BannerComponent element tag as mentioned.

braughtg:
> This is what documentation looks like for the BannerComponent.

It looks like everything after "Example" is interpreted as code and is being syntax highlighted.  E.g. the `For` is colored because it thinks it is a `for loop`.  So maybe it makes sense to structure the entire example as code?  Then put in comments to explain the pieces?  

So start with a comment saying: `<!-- The component is added to the page with a tag like this... -->` (HTML style comment)

Then after the tag: `// The event hander would be ...` 

Or whatever makes sense in your example.

FutzMonitor:
> It looks like everything after "Example" is interpreted as code and is being syntax highlighted. E.g. the `For` is colored because it thinks it is a `for loop`. So maybe it makes sense to structure the entire example as code? Then put in comments to explain the pieces?
> 
> So start with a comment saying: `<!-- The component is added to the page with a tag like this... -->` (HTML style comment)
> 
> Then after the tag: `// The event hander would be ...`
> 
> Or whatever makes sense in your example.

Unfortunately only comments denoted with the double slash `//` don't get the colored code treatment. 
image
I tried the HTML comment but even the words inside get colored.

FutzMonitor:
This is what it looks like now: 
image

FutzMonitor:
I also thought about that edge case where a timeout is being used and is set to true but then set to false before the timeout expires. There's no way for that to happen internally within the component but the parent page can pass a false which means that the variable in charge of banner visibility is now set to false. In that case, I don't think we need to emit at all. Especially because the change is being watched so the child component would also be set to false once the parent changes its value as well.

braughtg:
> In that case, I don't think we need to emit at all

Good point. Agreed.

braughtg:
> This is what it looks like now:

Does it look okay if the HTML element is not in a comment?  Can it be syntax highlighed in a reasonable way?

FutzMonitor:
> Does it look okay if the HTML element is not in a comment? Can it be syntax highlighed in a reasonable way?

This is what it looks like: 
image

braughtg:
> This is what it looks like:

Better!  Maybe just put a few line breaks in the html element, indent and align the attributes to make it easier to read than one long line and should be good to go.

FutzMonitor:
> Better! Maybe just put a few line breaks in the html element, indent and align the attributes to make it easier to read than one long line and should be good to go.

Something like this? 
image

Mind the space between the '@' and 'banner-hidden'. If they're right next to each other JSDoc thinks it's part of the syntax and doesn't load anything after it.

braughtg:
> Mind the space between the '@' and 'banner-hidden'. If they're right next to each other JSDoc thinks it's part of the syntax and doesn't load anything after it.

This looks great.  Though I do worry about that space.  I wonder if it would break (not assign the event handler) if copied into a Vue page?  I also wonder if there is a way in JSDoc to escape the @ so that it can be used in an example.  

Here's something that might help: https://snyk.io/advisor/npm-package/jsdoc-escape-at?utm_medium=referral&utm_source=yarn&utm_campaign=snyk-widget

FutzMonitor:
This is what it looks like with the npm package installed.

FutzMonitor:
With the latest commit, this PR also resolves issue #635. If there is going to be a new npm package included in the dev image that requires the image on Docker Hub to be updated then might as well tackle that small issue as well.

braughtg:
This all looks good now.  I like it! Well done.

I will try to look at building the new image and merging this over the weekend.

braughtg (approved):

PR #627, opened 2023-03-03
Surpressing Standard Name Error

### LINKED ISSUE #565: Standard Class Not Defined $name
## Current Behavior ##
1. Bring up FarmData2 using the `./fd2-up.bash` command. 
2. Enter the following URLs: `http://localhost/farm/fd2-field-kit` or `http://localhost/farm/fd2-barn-kit`
3. You should be treated to the login screen with an error banner at the top that looks like the following: 

image
*This is for the field kit URL. Attempting the barn kit URL changes the field to barn. It's the same error*

image
*What the page looks like as a whole*
## Desired Behavior ##
Because you're logged out and attempting to access a page without being logged in, those user names inside of the Field Kit [module](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/fd2_field_kit/fd2_field_kit.module#L92) and Barn Kit [module](https://github.com/DickinsonCollege/FarmData2/blob/main/farmdata2_modules/fd2_tabs/fd2_barn_kit/fd2_barn_kit.module#L101) are probably undefined since those pre-processes cannot run when you're not logged in. It would be best to simply modify the function that handles the pre-processing so that it doesn't print that error messages and just quits out when any of the names are undefined. This solution will require some understanding of PHP.

Resolving this issue will make the login page look like this when trying to access those URLs without being logged in: 

image

FutzMonitor:
__Pull Request Description__

Closes #565. Adds a simple simple `@` in front of the cmd command that maps the user ID's and names as a preprocessing function. This error would appear when users attempted to access a page without being logged into the page. The warning no longer appears due to this surpression and a comment explaining the `@` placement is now present in the module files. 

---
__Licensing Certification__

### DIFF
--- a/farmdata2/farmdata2_modules/fd2_barn_kit/fd2_barn_kit.module
+++ b/farmdata2/farmdata2_modules/fd2_barn_kit/fd2_barn_kit.module
@@ -97,6 +97,7 @@ function fd2_barn_kit_preprocess_page() {
     drupal_add_js('https://unpkg.com/dayjs@1.8.21/dayjs.min.js', 'external');
 
     // Define some useful variables in the JavaScript for page.
+    // the @ symbol surpresses the warning when a user attempts to access a page without being logged in
     global $user;
     @$cmd="var fd2UserID=".$user->uid."; var fd2UserName='".$user->name."';";
     drupal_add_js($cmd,'inline');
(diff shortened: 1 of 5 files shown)


braughtg:
This looks like a nice simple and elegant solution.  Can you provide a link to a reference in a comment here that I can review that will document what the @ symbol does in a PHP file like this?

FutzMonitor:
The documentation can be found [[here](https://www.php.net/manual/en/language.operators.errorcontrol.php)]

braughtg:
The `@` to suppress the error on these lines only seems like a reasonable approach for this particular issue. 

Please also check to see if there is an instance of this in the `fd2_config` module as well.

FutzMonitor:
@braughtg I added the `@` to the `fd2_config` module.

braughtg (approved):

PR #628, opened 2023-03-03
Filtering Area Based on Currently Edited Seeding Log

### LINKED ISSUE #585: Seeding Report: Filtering Area/Crop Name Arrays Depending on Seeind Log

FutzMonitor:
__Pull Request Description__

Closes #585. This PR aims to remove the user's ability to assign the incorrect area to seeding logs when they're being edited. For example, a direct seeding should **not** be allowed to be assigned a greenhouse area and a tray seeding should **not** be allowed to be assigned a field area. 

This pull request is **NOT** ready to be merged. 
There are a couple of things that need to be addressed before merging. 
- [ ] Conversation about implementation 

---
__Licensing Certification__

### DIFF
--- a/farmdata2/farmdata2_modules/fd2_barn_kit/seedingReport/seedingReport.html
+++ b/farmdata2/farmdata2_modules/fd2_barn_kit/seedingReport/seedingReport.html
@@ -27,7 +27,7 @@
                     </div>
                 </fieldset>
 
-                <custom-table data-cy="report-table" :columns="tableColumns" :rows='tableRows' csv-name="seedingReport_" @row-deleted="deleteRowLog" @row-edited='updateRow' @edit-clicked='disableFilters' @edit-canceled='enableFilters' :can-edit='!isMobile' :can-delete='!isMobile'></custom-table>
+                <custom-table data-cy="report-table" :columns="tableColumns" :rows='tableRows' csv-name="seedingReport_" @row-deleted="deleteRowLog" @row-edited='updateRow' @edit-clicked='updateRowEditID' @edit-canceled='enableFilters' :can-edit='!isMobile' :can-delete='!isMobile'></custom-table>
 
                 <br>
             </div>
@@ -96,6 +96,9 @@
                 reportVisible: false,
                 seedingLogs: [],
                 updateLog: [],
+                editingAreaFilter: [],
+                areaJSONList: [],
+                editRowID: null,
 
                 sessionToken: null,
                 selectedCrop: 'All',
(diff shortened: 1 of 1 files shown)


FutzMonitor:
There are a couple of issues that I have with my own implementation. However, it will hopefully serve as a starting point of conversation to resolve the overall issue. 

1. I have added a `getAllPages` call on line [864](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/FilteringAreaAndCrop/farmdata2/farmdata2_modules/fd2_barn_kit/seedingReport/seedingReport.html#L864) which fetches all the area JSON objects and saves them into a new Vue data variable called `areaJSONList` on line [100](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/FilteringAreaAndCrop/farmdata2/farmdata2_modules/fd2_barn_kit/seedingReport/seedingReport.html#L100). I made this decision because the computed function responsible for filling the `CustomTableComponent` with the areas only has the area names. It doesn't not have any other information (to my knowledge) that can help in distinguishing an area between being part of the greenhouse or being part of the field. Consequently, using the Seeding Input page as example, I decided to get the JSON objects to help me distinguish between the two. 
2. The `updateRowEditID` (name change pending) on line [156](https://github.com/FarmData-2-Dev-Team-2022/FarmData2/blob/FilteringAreaAndCrop/farmdata2/farmdata2_modules/fd2_barn_kit/seedingReport/seedingReport.html#L156) is incredibly inefficient in my mind. It's not noticeable when I'm testing it but filtering through the lists every *single* time we can to figure out if the seeding log being edited is a tray or direct seeding is wasteful of resources. A potential alternative that I though of is that after getting our response from the new `getAllPages` line added to the page, we can filter the areas into 2 new Vue data variables called (something along the lines of) `trayAreas` and `fieldAreas` and then we simply the `updateRowEditID` to simply figure out the log is a direct or tray seeding. Afterwards, either a computed or another Vue data variable could be saved with the needed areas and then passed into the computed method for the `CustomTableComponent`. 
3. I don't really like that I had to remove the `disableFilters` call from the event handler inside the `CustomTableComponent` element tag in the Seeding Report. If someone could point me to some sort of resource that allows me to call both while still being able to pass the event's payload to `updateRowEditID` that would be much appreciated. I tried following some examples like [this](https://stackoverflow.com/questions/38744932/vue-how-do-i-call-multiple-functions-with-click) one but the event's payload wasn't going through somehow.

PR #680, opened 2023-10-13
Change default seeding type according to existing logs

### LINKED ISSUE #658: BarnKit Seeding Report Default for Seeding Type Filter
Currently when a Seeding Report is generated the Seeding Type displayed defaults to "All" even if there are no Direct Seedings or Tray Seedings.  Would it be more useful to have the report automatically detect what types of seedings actually appear in the report and then default to Direct Seedings, Tray Seedings or All?

Mikek16:
__Pull Request Description__
This PR changes the `selected-val` of the Seeding Type Dropdown from 'All' to the `getDefaultSeedingSelection` function that returns the appropriate seeding type for the available logs.
Closes #658 
<Add description>

---
__Licensing Certification__

### DIFF
--- a/farmdata2/farmdata2_modules/fd2_barn_kit/seedingReport/seedingReport.html
+++ b/farmdata2/farmdata2_modules/fd2_barn_kit/seedingReport/seedingReport.html
@@ -18,7 +18,7 @@
                 <fieldset :disabled="rowBeingEdited" data-cy="filters-panel" class="center-block panel panel-default" style="width: 75%;">
                     <legend class="panel-heading" style="background-color: #d62a17; color: white;">Filters</legend>
                     <div :style=" !isMobile ? 'display: flex;width: 100%; padding: 7px' : 'display: grid; width: 100%'">
-                        <dropdown-with-all data-cy="seeding-type-dropdown" class="center-block" :dropdown-list='seedingTypeList' includes-all selected-val='All' @selection-changed='seedingChange' id="seedingType">Seeding Type:</dropdown-with-all>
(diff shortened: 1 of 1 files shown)


braughtg:
@FutzMonitor Would you have any interest in reviewing this PR? Let me know in chat on Zulip and I can advise.

braughtg:
@Mikek16 Thanks for this fix.  Please add a description of the changes to the PR description above and change the "Addresses" link to be "Fixes" or "Closes".  Thanks!

braughtg:
@Mikek16, @FutzMonitor has agreed to review this PR.  You should have some feedback later this week.

FutzMonitor (changes requested):
This looks good to me. The basic functionality is to take the pre-existing `seedingTypeList` computed by Vue and checking to see which seeding type isn't present. If tray seeding isn't present, then we can safely assume that only direct seeding logs are present and vice versa for tray seeding logs. 

It does make me question though if that's the case if we should add a dropdown that doesn't include all or simply disable the dropdown if there are no other log options. What're your thoughts @braughtg 

Other than that, there are some minor indentation errors I would like you to correct @Mikek16

FutzMonitor (approved):

braughtg:
> It does make me question though if that's the case if we should add a dropdown that doesn't include all or simply disable the dropdown if there are no other log options. What're your thoughts @braughtg

If I understand correctly this would be the case if for example there is a search that has both types of seeding, but then the user filters by crop or area, and as a result there is only one type of seeding. In that case, it does make sense to me that the "All" option should not appear in the drop down.  And since the other type of seeding also should not appear in the dropdown then you might also want to just disable the dropdown (as suggested).  

I think this could get complicated though.  The special case where someone filters for "Tray Seedings" would result in there only being tray seedings.  However, we would still need to have All and Direct Seedings in the dropdown.  I'm sure this can be done but it would need to be handled carefully.  

If @Mikek16 or @FutzMonitor want to open a new issue for this (including note of the above special case), that would be a nice contribution!  Thanks!

PR #681, opened 2023-11-08
Slight Optimization to Logging and Paging Endpoints and Parameterized SQL Queries 

WarpWing:
__Pull Request Description__
Made some slight formatting/optimization fixes to the FarmData2 Logging Endpoint. In addition, I also parameterized(in lieu of concatenation) SQL queries so that user input is treated as data, not as part of the SQL command itself. This prevents SQL injection and stops unauthorized commands from being run via the endpoint. Prepared statements utilize ? as placeholders for logType, startTime, endTime, startIndex, and limit. I also removed pagination and instead handled that in the LIMIT section of the query.

I'm currently putting this in a "DRAFT" pull request because I would like to test out some more behaviors before making this reviewable.

__Licensing Certification__


WarpWing:
In addition, just looked at /paging, and made some small refactoring including pagination via implementing LIMIT in the SQL statement.

braughtg:
Thanks for your work on this!  These look like nice updates.  

I don't see any obvious breaking changes, but as you work be sure to double check both manually and by running the tests in `farmdata2_api/cypress`.  The API is also used by functions in `farmdata2_modules/resources/FarmOSAPI.js` so be sure to run the tests in `FarmOSAPI.spec.js` as well.

WarpWing:
I forgot this fossil PR laying around but I tested these changes and they all look good on my end.

PR #688, opened 2023-12-03
Split FarmOSAPI test file

### LINKED ISSUE #651: Divide FarmOSAPI Tests into smaller files.
Break the tests here into smaller pieces with each being contained in its own file.  For example, the tests for getting maps should be broken out into `FarmOSAPI.maps.spec.js`. This will better allow developers to only run those tests that are being affected by current changes.  

The full `FarmOSAPI.spec.js` file should be examined and the exact division of the current tests into smaller files should be done based on the functionalities being tested.

nguyenbanhducA1K51:
__Pull Request Description__
close #651 

Create a test folder  and divide tests in `FarmOSAPI.js` into smaller chunks

---
__Licensing Certification__


braughtg (changes requested):
It seems like the same tests are appearing in multiple files.  Ideally, each test would appear only in the appropriate files.  E.g. the `Configuration.spec.js` file would only contain tests that test getting and setting the configuration values.  The other tests for things like `getRecord` or "get a non-existent record" woudl be in the `Record.spec.js` file that tests all of the functionality related to the basic record operations.

See if you can make a cleaner division of the tests into the files that you have here so that each test appears only once and that everything is still tested.

nguyenbanhducA1K51:
I have check again and remove duplicate testing section. Also I discovered that for the test `Test on a request with multiple pages `, sometimes it fails ( it passed on other running with total time around 5999ms ), and here is the screenshot. I think it need to be in another issue for future resolve. 
Screenshot 2023-12-14 191711
Screenshot 2023-12-14 191800

braughtg:
Tests run fine for me.  The timeout issue will be machine specific so individuals can increase it as necessary.

PR #690, opened 2023-12-03
add user role/permission to module fd2_example

### LINKED ISSUE #447: Accessing User Roles
When creating and displaying UI elements it will be necessary to know what roles a user has in the FarmOS system (e.g. Admin, Farm Manager, Farm Worker, Farm Viewer).  This will allow the UI to display only the controls that the current user has permission to use (e.g. Edit / Delete).

nguyenbanhducA1K51:
__Pull Request Description__
close #447 
<Add description>
Add role/permission to login user, currently it must be  imported in each module the same way like uid and name property. There could be a way to refactor so that every modules can have a common file to import from. Image below is from sub tab `fd2_example`, and I have remove that `p` tag in the commit, but the usage is straightforward.
Screenshot 2023-12-03 024447

---
__Licensing Certification__

### DIFF
--- a/farmdata2/farmdata2_modules/fd2_example/fd2_example.module
+++ b/farmdata2/farmdata2_modules/fd2_example/fd2_example.module
@@ -133,7 +133,22 @@ function fd2_example_preprocess_page() {
     // Define some useful variables in the JavaScript for page.
     // the @ symbol surpresses the warning when a user attempts to access a page without being logged in
     global $user;
-    @$cmd="var fd2UserID=".$user->uid."; var fd2UserName='".$user->name."';";
+
+
+   $account = user_load($user->uid);
+    // Map from roleID to role name
+    $roleMap = user_roles();
+    // Get the roles assigned to the user.
+    $roles = array();
+    foreach ($account->roles as $role => $roleName) {
+      $roles[] = $roleMap[$role];
+    }
+
+    @$cmd="var fd2UserID=".$user->uid."; var fd2UserName='".$user->name."'; ;";
     drupal_add_js($cmd,'inline');
+    $roleCMD = "var fd2Role = " . json_encode($roles) . ";"; // JSON encode roles array
+    drupal_add_js($roleCMD, 'inline');
+
+
   }
 };


braughtg:
Having the demonstration of how to access the roles in the example would be great.  Can you add those to the API example page again and include that in this PR?  Place the displayed roles just below the username.

nguyenbanhducA1K51:
I have add illustration in latest commit, and here is the image
Screenshot 2023-12-11 022752
Screenshot 2023-12-11 022723

farmdata2/farmdata2_modules/fd2_example/maps/maps.html
braughtg, line 1: This looks good, but ideally it should go into the "Vars" example tab instead of the "Maps" tab.  The "Vars" example illustrates the use of the variables that are injected into the page.  Including it here, makes it look like the role information is coming from the `IDToUserMap` which it is not. 

Please move the part that shows the roles to the "Vars" example tab instead of the "Maps" example tab.

braughtg (changes requested):
Nice work getting the roles added to the JS variables!  Well done.

I added a comment about moving the illustration from the "Maps" example to the "Vars" example where it fits more naturally.

Also, there is a test file `fd2_example/vars.uid.spec.js` that tests that the JS variables for the user ID and username work correctly.  Please extend the tests in that file to include assertions (`.should`) about the roles that the user has.

nguyenbanhducA1K51:
I have move it to var subtab and add some test `fd2_example/vars.uid.spec.js`
Screenshot 2023-12-14 162331

braughtg:
This all appears to be working well within the `fd2_example` module and the variables seem to have been added to the `fd2_barn_kit` module.  To complete the work on this issue you'll need to add the variables to the `fd2_field_kit` and add tests to the `cypress/fd2vars.spec.js` in the `fd2_barn_kit` and `fd2_field_kit` modules  (like the ones you added in the `fd2_examples` module) to be sure that the variables are functioning correctly in those modules as well.

